PR #97, opened 2024-11-28
docs: add new-cell guide

EwecaBcD:


guide/src/guide/new-cell.md
```diff
@@ -1 +1,221 @@
-# Write a New Cell (TODO)
+# Write a New Cell
+
+Following this guide, you can write a custom cell yourself.
+
+## Basic Structure
```
BobAnkh, line 5: It is important to convey the core concepts of a cell to the developers reading this guide. What is a cell? How is it composed? Which traits should be implemented, and what logic should be implemented in what methods?

We should first state these high-level pictures and then on the details of how to implement a illustrating cell (i.e., LossCell)

guide/src/guide/new-cell.md
```diff
@@ -1 +1,221 @@
...
+
+We have some necessary structs and traits which you must impl for your own cell. We take a cell which loses one packet in every `n` packets (`n` can be set by users freely) as example to explain basic structure of a custom cell.
```
BobAnkh, line 7: `impl` to `implement`. Abbreviations should be avoided in the guide (you can use abbreviations in documents or comments).

You can use `drop` instead of `lose` for verb and `Lost` or `Drop` for the name of the cell.

guide/src/guide/new-cell.md
```diff
@@ -1 +1,221 @@
...
+
+- `CellIngress`
```
BobAnkh, line 9: We should construct the outline with the name of the traits. 
We should write our guide around the concept of `trait objects`.
You can make them as 3rd level title:

```txt
### Ingress
```

As for `Ingress`, I can give you some example sentences for reference:

---

We should implement an `Ingress` to receive packets, which are dequeued by other cells and forwarded by the rattan core. The `Ingress` serves as the entry of our cell. Since we allow multiple cells to send packets to a single cell from different threads, `Ingress` must implement the `Clone` trait and can only have an immutable reference to itself. However, it's important to note that mutable operations on the packets are possible within the `enqueue` method of `Ingress`.

guide/src/guide/new-cell.md
```diff
@@ -1 +1,221 @@
-# Write a New Cell (TODO)
+# Write a New Cell
```
BobAnkh, line 1: We should tell the readers why and how they can implement a new cell for the whole guide. That is, why we should implement certain traits and how it affects the implementation, and what we can do and what we cannot do inside the trait.

guide/src/guide/new-cell.md
```diff
@@ -1 +1,221 @@
...
+              control_interface: Arc::new(LosePacketCellControlInterface { loss_interval }),
+          })
+      }
+  }
+  ```
+
+## Add Configuration to Your New Cell
+
+- TODO: How to add configuration support to a new cell(In tests and in .toml files)
+
```
BobAnkh, line 210: We should state the premise of this section: if you want to integrate the cell into the rattan official repository to the corresponding CLI tool.

BobAnkh (approved):

Issue #80, opened 2024-07-29
IO Safety violation runtime error in debug mode


un-lock-able:
When running in debug mode, the binary cannot start correctly but instead throws out runtime error and terminates itself after creating the veth pair. However, if we swich to release mode, the binary can successfully start and behaves as expected.

Terminal log to reproduce:
```
unlockable@theresa:~/repos/rattan$ cargo run --release -- --uplink-delay 10ms
    Finished `release` profile [optimized + debuginfo] target(s) in 0.09s
     Running `sudo -E target/release/rattan --uplink-delay 10ms`
root@theresa:~/repos/rattan# ping 10.2.1.1 -c 3
PING 10.2.1.1 (10.2.1.1) 56(84) bytes of data.
64 bytes from 10.2.1.1: icmp_seq=1 ttl=64 time=10.3 ms
64 bytes from 10.2.1.1: icmp_seq=2 ttl=64 time=10.4 ms
64 bytes from 10.2.1.1: icmp_seq=3 ttl=64 time=10.7 ms

--- 10.2.1.1 ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2003ms
rtt min/avg/max/mdev = 10.259/10.444/10.689/0.180 ms
root@theresa:~/repos/rattan# exit
exit
unlockable@theresa:~/repos/rattan$ cargo run -- --uplink-delay 10ms
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.15s
     Running `sudo -E target/debug/rattan --uplink-delay 10ms`
fatal runtime error: IO Safety violation: owned file descriptor already closed
Aborted
unlockable@theresa:~/repos/rattan$
```

BobAnkh:
Provide some references to fix this issue:

- [alacritty/alacritty#7996](https://github.com/alacritty/alacritty/pull/7996)
- [zed-industries/zed#12114](https://github.com/zed-industries/zed/issues/12114)

un-lock-able:
I dig down some more and I suspect that now there are multiple places involving `NetNS` are all causing the problem.

In `rattan-core/src/metal/veth.rs`, inside `VethPairBuilder::build_impl()`, adding a piece of code of 
```rust
let current_env = NetNS::current()?;
drop(current_env);
```
would immediately trigger this error message. I now suspect that the `NetNS::drop()` function does not handle the fd it owns gracefully and the fd could be closed multiple times.

[...]

PR #99, opened 2024-12-19
fix(netns): avoid double-close persistent file

BobAnkh:
Resolves: #80

### DIFF
--- a/rattan-core/src/metal/netns.rs
+++ b/rattan-core/src/metal/netns.rs
@@ -305,10 +305,6 @@ impl<E: Env> NetNs<E> {
 
 impl<E: Env> Drop for NetNs<E> {
     fn drop(&mut self) {
-        let fd = self.file.as_raw_fd();
-        if let Err(e) = nix::unistd::close(fd).map_err(NsError::CloseNsError) {
-            error!("Failed to close netns: {}", e);
-        }
         if let Err(e) = self.env.clone().remove(self) {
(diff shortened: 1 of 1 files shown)


BobAnkh:
The fix seems quite simple. The `remove` method of `DefaultEnv` will always `remove_file` the path of the persistent file created by `NetNs::new`. It won't remove the file created by `NetNs::current`. However, the file owned by `NetNs::current` will be auto-closed in the drop function. There might be no need to explicitly close the persistent file before we let env remove that netns. See if @minhuw has some words on this.

rattan-core/src/metal/netns.rs
```diff
@@ -305,10 +305,10 @@ impl<E: Env> NetNs<E> {
...
 impl<E: Env> Drop for NetNs<E> {
     fn drop(&mut self) {
-        let fd = self.file.as_raw_fd();
-        if let Err(e) = nix::unistd::close(fd).map_err(NsError::CloseNsError) {
-            error!("Failed to close netns: {}", e);
-        }
+        // let fd = self.file.as_raw_fd();
+        // if let Err(e) = nix::unistd::close(fd).map_err(NsError::CloseNsError) {
+        //     error!("Failed to close netns: {}", e);
+        // }
```
minhuw, line 311: LGTM. Better delete it directly.

minhuw (approved):
LGTM.

PR #122, opened 2025-06-04
Adding delay cells that can reorder packets

maxime-bruno:
Adds a new cell `ReorderDelayCell`.

This cell takes a delay generator. The packets are delayed in a way (priority queue) that allows potential reordering.

I added support for :
  - Constant delay
  - Normal Law based delay
  - Log-Normal Law based delay

It can be a potential solution for issue #32 

I also added benchmarks to test the bandwidth, delay, and reorder delay cell.


rattan-core/src/cells/reorder_delay/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+    }
+}
+
+pub struct ReorderDelayCellEgress<P, D>
+where
+    P: Packet,
+    D: DelayGenerator,
+{
+    egress: mpsc::UnboundedReceiver<P>,
+    delay: D,
```
BobAnkh, line 58: rename it to reorder_trace

rattan-core/src/cells/reorder_delay/delay.rs
```diff
@@ -0,0 +1,65 @@
+use rand_distr::Distribution;
+use tokio::time::Duration;
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+/// A delay generator that can be used in the reorder delay cell.
+pub trait DelayGenerator {
```
BobAnkh, line 8: Please rename it to ReorderDelayTrace and rename the method to `next_delay`

BobAnkh (changes requested):
@maxime-bruno We really appreciate your contributions and your overall design! However, there remain some points to be discussed besides the changes requested:

1. Naming conventions: Could you please rename the `reorder_delay` or `ReorderDelay` just to `reorder` or `Reorder` for simplicity? For our design choice to choose delay as our trace definition, we can add some words in documentations. Additionally, could you please rename the `DelayGenerator` trait to `ReorderTrace` and its method to `next_delay` to ensure compatibility with other trace traits in our crate?
2. Could you please remove the file `.pre-commit-config.yaml`?
3. For the `DelayGenerator` (or `ReorderTrace` as I expected) trait (and its specific models, normal or log normal), I will move it out of this crate to crate `netem-trace` in the near future, as we did with other cells for decoupling. You can also do this directly by sending a PR to that codebase if you prefer.
4. Could you please add a demo configuration section to the file `config.example.toml`?
5. Could you please add a test for correctness verification to `tests/integration`, similar to what we did for other cells?
6. For the definition of `ReorderCellBuildConfig`, I do think we can change it to be similar to [`DelayReplayCellBuildConfig`](https://github.com/stack-rs/rattan/blob/d38b9e6865b486f776db0451dfe1d826246f2769/rattan-core/src/config/delay.rs#L29). In this way, we can easily extend reorder trace with more models beside normal or log normal in other crates (i.e., `netem-trace`) without modifying rattan. If you want to preserve some of them for simpler configuration, you can also add a new `ReorderReplayCellBuildConfig` struct.

rattan-core/benches/cells/bandwidth.rs
```diff
@@ -0,0 +1,110 @@
...
+        40..50 => format!("{:.3}TiBps", bytes_ps as f64 / 2_u64.pow(40) as f64),
+        _ => format!("{:.3}PiBps", bytes_ps as f64 / 2_u64.pow(50) as f64),
+    }
+}
+
+pub fn run(group: &mut BenchmarkGroup<WallTime>, handle: &Handle) {
+    for bandwidth in [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096]
+        .into_iter()
+        .map(|mbps| Bandwidth::from_bps(mbps * 1024 * 1024))
+    {
```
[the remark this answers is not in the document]
maxime-bruno: I usually use SI units too, but `criterion` is outputting the throughput in `Mi` units, that is why I chose `Mi`, to easily see the differences between the expected throughput and the real one.

BobAnkh:
Hi,

For your concerns:

> I haven't understood how replay cells are supposed to work, that's why I didn't implement one. Because I guess, at some point the delay should then be deterministic, which is not the case here

I can explain more details to you. For all such cells, it will replay a so-called `trace` (e.g., `BwTrace`, `DelayTrace`, etc.), and in our case `DelayGenerator` (or `ReorderTrace` as I expected). The `trace` should give a deterministic sequence on a given input, but this doesn't mean the sequence cannot be randomly generated. The output sequence can be subject to some distributions. For example, for bandwidth replay cell, it can replay bandwidth generated from a normal distribution ([NormalizedBw](https://github.com/stack-rs/netem-trace/blob/f1a389f1114b9efc716497c28887cd94c4b114bb/src/model/bw.rs#L165) and [BwReplayCell](https://github.com/stack-rs/rattan/blob/d38b9e6865b486f776db0451dfe1d826246f2769/rattan-core/src/cells/bandwidth/mod.rs#L692)). I forget to mention this in my reviews: you have to fix the seed or provide option to set the seed for the random-generated distribution. You can see how I implement this in [NormalizedBw](https://github.com/stack-rs/netem-trace/blob/f1a389f1114b9efc716497c28887cd94c4b114bb/src/model/bw.rs#L165).

> The name ReorderDelayCell was due to the fact that it is mostly a delay cell (that can perform reordering), and technically could replace the DelayCell using just a constant delay generator. And one may want to implement a cell that just really reorders packets in the future

While it may technically replace the `DelayCell`, it cannot replace the complicated `DelayReplayCell`. But I agree with you that we can leave its name as `ReorderDelayCell`, as we may have other reordering implementations in the future. But for the folder or file name, I think we can shrink it to reorder.

Let me know if you have any further thoughts or concerns.

BobAnkh:
In detail, the replay cell is a cell that contains a trait object, which produces a sequence. As you can see the `Option<Box<dyn BwTraceConfig>>` in the [BwReplayCellConfig](https://github.com/stack-rs/rattan/blob/d38b9e6865b486f776db0451dfe1d826246f2769/rattan-core/src/cells/bandwidth/mod.rs#L599):

```rust
pub struct BwReplayCellConfig<P, Q>
where
    P: Packet,
    Q: PacketQueue<P>,
{
    pub trace_config: Option<Box<dyn BwTraceConfig>>,
    pub queue_config: Option<Q::Config>,
    pub bw_type: Option<BwType>,
}
```

or the `trace: Box<dyn BwTrace>` in [BwReplayCellEgress](https://github.com/stack-rs/rattan/blob/d38b9e6865b486f776db0451dfe1d826246f2769/rattan-core/src/cells/bandwidth/mod.rs#L391):

```rust
pub struct BwReplayCellEgress<P, Q>
where
    P: Packet,
    Q: PacketQueue<P>,
{
    egress: mpsc::UnboundedReceiver<P>,
    bw_type: BwType,
    trace: Box<dyn BwTrace>,
    packet_queue: Q,
    current_bandwidth: Bandwidth,
    next_available: Instant,
    next_change: Instant,
    config_rx: mpsc::UnboundedReceiver<BwReplayCellConfig<P, Q>>,
    send_timer: Timer,
    change_timer: Timer,
    state: AtomicI32,
}
```

So that one we have a model implementing `BwTrace` (e.g., [NormalizedBw](https://github.com/stack-rs/netem-trace/blob/f1a389f1114b9efc716497c28887cd94c4b114bb/src/model/bw.rs#L165), similar to your `NormalLawDelayGenerator`. We can even have a more complex combination of models), we can use the same cell to replay them,  whether it is Normal or LogNormal.

For an example usage, see [tests/integration/bandwidth.rs#L773](https://github.com/stack-rs/rattan/blob/d38b9e6865b486f776db0451dfe1d826246f2769/rattan-core/tests/integration/bandwidth.rs#L773):

```rust
let trace = RepeatedBwPatternConfig::new()
    .pattern(vec![Box::new(StaticBwConfig {
        bw: Some(Bandwidth::from_mbps(1)),
        duration: Some(Duration::from_secs(10)),
    }) as Box<dyn BwTraceConfig>])
    .build();
BwReplayCell::new(
    Box::new(trace) as Box<dyn BwTrace>,
    DropTailQueue::new(DropTailQueueConfig::new(100, None, BwType::default())),
    None,
)
```

maxime-bruno:
So a proper implementation of what I called DelayGenerator would be 
```rust
/// This is a trait that represents a trace of per-packet delays.
///
/// The trace is a sequence of `delay`.
/// The delay describes how long the packet is delayed when going through.
///
/// For example, if the sequence is [10ms, 20ms, 30ms],
/// then the delay will be 10ms for the first packet, then 20ms for second, then 30ms for third.
///
/// The next_delay function either returns **the next delay**
/// in the sequence, or **None** if the trace goes to end.
pub trait DelayPerPacketTrace: Send {
    fn next_delay(&mut self) -> Option<Delay>;
}
```

`DelayPerPacket` seems more pertinent than `Reorder` and follows the guidelines (if I'm right)

BobAnkh:
Yes, I agree entirely with your design and implementation. It's just a matter of whether to call it `DelayPerPacketTrace` or `ReorderDelayTrace`, which you can decide on your own.

config.example.toml
```diff
@@ -245,6 +245,11 @@ type = "Loss"
 # Default: []
 pattern = [0.1, 0.2]
 
+# Delay Per-Packet Cell Example
+[cells.down_3]
+type = "DelayPerPacket"
+
+FromFile = "./assets/delay-per-packet.toml"
```
BobAnkh, line 252: Please use the name `trace` instead of `FromFile` as to be compatible with other cells. Also we prefer snake_case in config file.

rattan-core/tests/integration/per_packet/delay.rs
```diff
@@ -0,0 +1,136 @@
+use netem_trace::model::StaticDelayPerPacketConfig;
+use netem_trace::Delay;
+use rattan_core::cells::{per_packet::delay::DelayPerPacketCellConfig, StdPacket};
+/// This test need to be run as root (CAP_NET_ADMIN, CAP_SYS_ADMIN and CAP_SYS_RAW)
+/// RUST_LOG=info CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER='sudo -E' cargo test delay --all-features -- --nocapture
```
BobAnkh, line 5: Could you please move this to the top of the file?

rattan-core/src/config/per_packet.rs
```diff
@@ -0,0 +1,59 @@
...
+    Figment,
+};
+use netem_trace::model::DelayPerPacketTraceConfig;
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Clone, Debug)]
+pub enum DelayPerPacketCellBuildConfig {
+    FromFile(std::path::PathBuf),
```
BobAnkh, line 21: Should this be changed to `Trace` in response to my requested changes in `config.example.toml`? If so, we change `FromConfig` to `config` as well. Also, please add a comment to `config.example.toml` to should how to use config to set the cell instead of trace file.

BobAnkh (changes requested):
Most of the changes are great! In addition to the requested changes, here are some requirements regarding cleaning up the commit history that you can refer to and modify before the final merge:

- Can you avoid modifying the svg figures inside assets? This makes the PR a little messy.
- And could you seperate the style format of files（e.g., devcontainer.json, .sockerignore, .github/workflows, ci/images) into a single commit in this PR so that I can easily maintain a clear git history?
- If you want to squash the commits in the PR (into two, one main feature and the other formatting) according to our commit conventions before merging, you can wait for my final commit message requirements. If you prefer me to handle it directly, you can change the target branch to the dev branch I newly created, and I will rebase after merging this PR.

maxime-bruno:
Seems doable 

The modification to assets and things like that are just trailing whitespace and missing new lines, I'll make a clean commit for that 

For squashing the commits, I'll let you handle it to be sure it has the names you want, or if you need me to separates commits I'll do it

BobAnkh:
> The modification to assets and things like that are just trailing whitespace and missing new lines, I'll make a clean commit for that

Then it's fine to include them in this PR as a separate commit.

> For squashing the commits, I'll let you handle it to be sure it has the names you want, or if you need me to split commits, for features separation, I'll do it

You can commit whatever you want for now, and we will handle this just before the merging process. For the final history, I expect there to be two squashed commits, one for all the features and the other for the style format (i.e., the trailing whitespaces or new lines, etc.). I will provide you with the commit message I need for the two commits before the merge, and you can assist me in rebasing the commit history so that I can merge them directly into the main branch.

maxime-bruno:
Hi,

Normally everything is done, I just have to push a PR for `netem-trace` once you are ok with everything and then it should be able to be merged

BobAnkh (approved):
LGTM

BobAnkh (approved):

BobAnkh:
> I did not test this part and just noticed this issue
> 
> The fact that the `DelayPerPacketCellBuildConfig` is an enum, the variant starts with a capital letter and that the `TOML` keyword is not capitalized does not work

Ok, can you send me a new PR to fix this issue? I think we may have some serde option to change this behavior (maybe something like snake_case, I think)?

PR #126, opened 2025-06-18
Spy cell

maxime-bruno:
Adds a new cell `SpyCell`.

Unlike the `packet_dump` feature, this cell allows the logging of the packet at a specific point in the network.

This is designed to have any type of writer (`std::io::Write`).

The `SpyCellBuildConfig` can take either:
  - a path: this creates the file, useful because the other needs `Clone` and `File` does not implement `Clone`
  - a `Box<dyn Write + Clone>`: can be used with Cursors, to write in memory

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -251,6 +251,12 @@ type = "DelayPerPacket"
 
 trace = "./assets/delay-per-packet.toml"
 
+# Spy Cell Example
+[cells.spy]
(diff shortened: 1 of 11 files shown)


BobAnkh (approved):
LGTM

BobAnkh:
Awesome work! Thank you for your contributions!

PR #127, opened 2025-06-19
Correcting some issues with serde

maxime-bruno:
I noticed that integration tests could not be run without the `serde` feature, and with it I did not noticed some issues with `DelayPerPacketCell`.

So I modified the test to work without `serde` features (and correct some issues)

### DIFF
--- a/rattan-core/src/config/delay.rs
+++ b/rattan-core/src/config/delay.rs
@@ -49,8 +49,8 @@ impl DelayReplayCellBuildConfig {
             }
         }
         Err(Error::ConfigError(format!(
-            "Unknown trace file format: {:?}",
-            file_path
+            "Unknown trace file format: {}",
+            file_path.display()
         )))
     }
 
--- a/rattan-core/src/control/mod.rs
+++ b/rattan-core/src/control/mod.rs
@@ -42,12 +42,12 @@ pub enum RattanOp {
 impl Debug for RattanOp {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            RattanOp::SendNotify(notify) => write!(f, "SendNotify({:?})", notify),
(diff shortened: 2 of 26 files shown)


rattan-core/src/config/delay.rs
```diff
@@ -49,8 +49,8 @@ impl DelayReplayCellBuildConfig {
             }
         }
         Err(Error::ConfigError(format!(
-            "Unknown trace file format: {:?}",
-            file_path
+            "Unknown trace file format: {}",
+            file_path.display()
```
Centaurus99, line 53: Lints like `clippy::unnecessary_debug_formatting` appear to be allowed by default yet.
I agree with this fix, but I'm also curious about what flags you used to run Clippy—this might help us decide whether to include them in CI to ensure consistent formatting.

rattan-core/src/control/mod.rs
Centaurus99, line 1: The `clippy::uninlined_format_args` lint will be changed back to a warning by default in Rust 1.88.0 (ref: <https://github.com/rust-lang/rust-clippy/pull/14160>). @BobAnkh 
Thanks for the fix! I was a bit confused during the review, so just leaving a note here for future reference.

Centaurus99 (changes requested):
Apologies for the delayed review, and thank you for the careful fixes!

Would it be possible to squash the commits into two—one for serde-related changes and one for linting fixes? It would make rebasing onto the main branch more cleaner. Thanks a lot!

rattan-core/src/config/delay.rs
maxime-bruno: I used `cargo +nightly  clippy --all-features --workspace --tests -- -Dwarnings` it showed several issue when you have a placeholder with a simple var to fill it 

In this case it complained about that, but noticed it would be better to use the display function

maxime-bruno:
I'll try to do the rework you are asking,
Sadly, it will have to wait (one week or two), I have other deadlines first

Issue #128, opened 2025-06-26
The log file doesn't take the endianness into account.


ympcMark:
When generating a log file, the use of little-endian output causes the byte order of the log file content to be inconsistent with the description in the document. The error code is at https://github.com/stack-rs/rattan/blob/b9f83010fc58c85d1389bcaa65796e539a2314cb/rattan-core/src/radix/log.rs#L125

PR #129, opened 2025-06-26
update log.rs

ympcMark:
Fix the endianness issue.


BobAnkh:
I will conduct a final code review once everything has been approved by @Centaurus99. Please ping me when all work is done.

Cargo.toml
```diff
@@ -52,6 +52,7 @@ http = ["rattan-core/http"]
 packet-dump = ["rattan-core/packet-dump"]
 
 [dependencies]
+byteorder = "1.5.0"
```
BobAnkh, line 55: Move it to workspace dependencies, as this is required by two crates.

rattan-core/src/radix/log.rs
```diff
@@ -40,6 +41,7 @@ pub struct FlowEntry {
 // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 // |          ip.checksum          |   tcp.flags   |    padding    |
 // +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
```
BobAnkh, line 44: Remove this line. The above comment should be closely attached to the TCPLogEntry

rattan-core/src/radix/log.rs
```diff
@@ -75,13 +77,12 @@ impl TCPLogEntry {
...
-                self as *const Self as *const u8,
-                std::mem::size_of::<Self>(),
-            )
-        }
+    pub fn as_bytes(&self) -> Vec<u8> {
+        let mut buf = Vec::with_capacity(32);
+        buf.extend_from_slice(&self.header.as_bytes());
+        buf.extend_from_slice(&self.general_pkt_entry.as_bytes());
+        buf.extend_from_slice(&self.tcp_entry.as_bytes());
+        buf
```
BobAnkh, line 85: I think this is not `as_bytes`. I wonder if we can really achieve `as_bytes`? If we can't, we should rename this to `to_bytes` or `get_bytes`.

Or can we define it just as [u8;n] and convert part of this inner buffer to the concrete type?

cc @Centaurus99 what is your idea?

rattan-core/src/radix/log.rs
```diff
@@ -225,6 +203,22 @@ pub struct TCPProtocolEntry {
...
+        buf[10..14].copy_from_slice(&self.ack.to_be_bytes());
+        buf[14..16].copy_from_slice(&self.ip_id.to_be_bytes());
+        buf[16..18].copy_from_slice(&self.ip_frag_offset.to_be_bytes());
+        buf[18..20].copy_from_slice(&self.checksum.to_be_bytes());
+        buf[20] = self.flags;
+        buf[21] = self.padding;
+        buf
+    }
+}
+
```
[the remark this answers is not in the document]
BobAnkh: Now I think we may use byteorder

rattan-core/src/radix/log.rs
```diff
@@ -270,16 +282,126 @@ impl Default for ProtocolHeader {
     }
 }
 
+More actions
```
BobAnkh, line 285: What is this?

BobAnkh (changes requested):

rattan-core/src/radix/log.rs
Centaurus99: I agree. The name `as_bytes` can be misleading, as it suggests returning a byte slice without copying. I prefer renaming it to `to_bytes` here, as well as in other places where `Vec<u8>` or `[u8; n]` is returned.

However, I now realize that the current implementation of `to_bytes` introduces additional copying for each layer of nesting. Perhaps we should implement a function like `write_to(&self, buf: &mut [u8])` instead, so copying only happens when writing to the buffer at the very end.

In fact, regarding custom packet structures, there are already many existing crates we could use, such as [packed_struct](https://crates.io/crates/packed_struct), [deku](https://crates.io/crates/deku), [binrw](https://crates.io/crates/binrw), and so on. Perhaps we should adopt one of these to avoid manually implementing these tedious parts and make it easier to further extend `LogEntry` in the future. @BobAnkh

BobAnkh: For current implementations, as long as the serialized results can still have the same size, I'm OK with them. If they will introduce other overheads, then we should implement such features ourselves.

BobAnkh: Also the output data format by these crates should be the same as defined in our protocol.

rattan-core/src/radix/log.rs
ympcMark: OK

rattan-core/src/radix/log.rs
ympcMark: sorry for this code, already deleted

rattan-core/src/radix/log.rs
Centaurus99: Using crates like [deku](https://crates.io/crates/deku), [binrw](https://crates.io/crates/binrw) seems to require us to expand the packed structs, which will result in additional struct size, causing the size of `TCPLogEntry` to exceed the carefully designed 32 bytes (within a cache line size). 
Therefore, I finally decided to implement it manually.

Centaurus99:
Finally, I realized that using big-endian format to store the packet log may not be a good choice. Endian conversion always comes with overhead. We would better write the file directly using the system's endian format (assumed as little-endian for convenience). Resolving #128 only requires rearranging the bit range.
Therefore, I will close this PR and continue in #133.

Issue #39, opened 2024-04-16
Drop privilege when spawn user threads


Centaurus99:
Maybe we can use <https://crates.io/crates/privdrop>.

PR #130, opened 2025-07-07
Not creating the spycell file when it exists

maxime-bruno:
This modification enables the opening and truncation of the spycell log file, even when the file is not owned by rattan user (currently root).

It is useful for the case like #39, when the log file is created by a user but rattan needs to edit it (it cannot create it as it exists but is not owned by the right user)

Maybe a log could be interesting to notify that the file already existed.

### DIFF
--- a/rattan-core/src/config/spy.rs
+++ b/rattan-core/src/config/spy.rs
@@ -1,4 +1,7 @@
-use std::ops::Deref as _;
+use std::{
+    fs::{File, OpenOptions},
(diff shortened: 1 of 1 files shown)


BobAnkh:
I would like to confirm the situation. Could you please explain or confirm in more detail? Currently, this modification should not have any changes, as the use of the `File::create` interface should be equivalent to the current logic? Or does this resolve some bug? Or is it just for preparing for certain cases in the future?

PR #138, opened 2025-07-30
fix: Remove unnecessary memset upon receiving af_packet.

Lethe10137:
When running 
```bash
cargo flamegraph -p rattan-core --root --example channel
```
, a weird flattop  of the libc funciton `__memset_avx2_unaligned_erms` was found when initializing the receiving buffer of a af_packet, which is totally unnecessary.

On my machine, the measurement of this example bitrate: 787.325101mbps  -> bitrate: 947.789548mbps.


BobAnkh:
Besides, you can just create new branches from this repo to open the PR, so that we can run every CI tests.

PR #143, opened 2025-09-08
Calibrate the start time of traces to first packet

Lethe10137:


BobAnkh (changes requested):
First, we shall still keep left and right in main.rs to not mess with the stdout (roll back to `Stdio::null()` instead of `Stdio::inherit()`). But I agree that we can add two arguments to enable inheritance individually.

Secondly, I think we should make it configurable of the starting behavior. Currently, I think we can just add a feature gate **or** another configurable option on cell initialization. We can default to enable the feature or set the option to FirstPacket. You can choose which plan you like to implement.

rattan-core/src/core.rs
```diff
@@ -271,33 +276,21 @@ where
...
+                                    if (rx_id.starts_with("left") || rx_id.starts_with("right")) && 
+                                       FIRST_PACKET_INSTANT.get().is_none() &&
+                                       FIRST_PACKET_INSTANT.set(tokio::time::Instant::now()).is_ok() {
+                                        if let Err(e) = notify_tx.exec(RattanOp::SendNotify(RattanNotify::FirstPacket)).await {
+                                            warn!(rx_id, tx_id, "Failed to send FirstPacket notify: {e:?}");
+                                        } else {
+                                            debug!(rx_id, tx_id, "Sent FirstPacket notify");
+                                        }
+                                    }
+
```
BobAnkh, line 293: Can you use `get_or_init` or `get_or_try_init` instead of a get and a set?

BobAnkh (changes requested):

rattan-core/src/core.rs
Lethe10137: However, neither `get_or_init` nor `get_or_try_init` tells whether a value was  successfully inserted.   Actually the .get() can be avoided, at the cost of calling the Instant::now() everytime.

BobAnkh: > However, neither get_or_init nor get_or_try_init tells whether a value was successfully inserted

Could you please explain why you need to know this?

PR #145, opened 2025-10-27
Timestamp

Lethe10137:
Rewrite:
- Delay(Replay)Cell
- TokenBucket

Modify:
- Bandwidth(Replay)Cell

Now a packet is delayed based on a "logical" timestamp which is marked on it when it entered rattan, and advanced by the cells above.

### DIFF
--- a/rattan-core/src/cells/delay.rs
+++ b/rattan-core/src/cells/delay.rs
@@ -55,7 +55,7 @@ where
     P: Packet,
 {
     egress: mpsc::UnboundedReceiver<P>,
-    delay: CurrentConfig<Delay>,
+    delay: Delay,
     config_rx: mpsc::UnboundedReceiver<DelayCellConfig>,
     timer: Timer,
     state: AtomicCellState,
@@ -69,7 +69,7 @@ where
     P: Packet + Send + Sync,
 {
     fn set_config(&mut self, config: DelayCellConfig) {
-        self.delay.update(config.delay, Instant::now());
+        self.delay = config.delay;
     }
 }
 
@@ -93,13 +93,7 @@ where
         let timestamp = packet.get_timestamp();
 
         let send_time = loop {
-            // `get_current` returns None if and only if `self.delays` has never been updated
-            // since last reset.
-            let logical_send_time = if let Some(delay_time) = self.delay.get_current(timestamp) {
-                (timestamp + *delay_time).max(self.latest_egress_timestamp)
-            } else {
-                timestamp.max(self.latest_egress_timestamp)
-            };
+            let logical_send_time = (timestamp + self.delay).max(self.latest_egress_timestamp);
 
             tokio::select! {
                 biased;
@@ -203,13 +197,11 @@ where
         let (config_tx, config_rx) = mpsc::unbounded_channel();
 
         let logical_time = *CALIBRATED_START_INSTANT.get_or_init(Instant::now);
-        let mut delay_setting = CurrentConfig::default();
-        delay_setting.update(delay, logical_time);
         Ok(DelayCell {
             ingress: Arc::new(DelayCellIngress { ingress: rx }),
             egress: DelayCellEgress {
                 egress: tx,
-                delay: delay_setting,
+                delay,
                 config_rx,
                 timer: Timer::new()?,
                 state: AtomicCellState::new(CellState::Drop),
@@ -582,8 +574,8 @@ mod tests {
         let received = received.unwrap();
         assert!(received.length() == 256);
 
-        assert_eq!(received.delay(), Duration::from_millis(10));
-        assert!((duration - 10.0).abs() <= DELAY_ACCURACY_TOLERANCE);
+        assert_eq!(received.delay(), Duration::from_millis(20));
+        assert!((duration - 20.0).abs() <= DELAY_ACCURACY_TOLERANCE);
 
         let start = Instant::now();
         let test_packet = TestPacket::<StdPacket>::with_timestamp(&[0; 256], start);
@@ -592,9 +584,10 @@ mod tests {
 
         // Wait for 15ms, then change the config back to 10ms
         std::thread::sleep(Duration::from_millis(15));
-        // Set after the timestamp of packets, so does not take effect.
         config_changer.set_config(DelayCellConfig::new(Duration::from_millis(10)))?;
 
+        // The expected behavior is that, the packet exits the cell almost immediately
+        // after the config change.
         let received = rt.block_on(async { egress.dequeue().await });
 
         let duration = start.elapsed().as_micros() as f64 / 1000.0;
@@ -604,9 +597,8 @@ mod tests {
         assert!(received.is_some());
         let received = received.unwrap();
         assert!(received.length() == 256);
-
-        assert_eq!(received.delay(), Duration::from_millis(20));
-        assert!((duration - 20.0).abs() <= DELAY_ACCURACY_TOLERANCE);
+        assert_eq!(received.delay(), Duration::from_millis(10));
+        assert!((duration - 15.0).abs() <= DELAY_ACCURACY_TOLERANCE);
 
         Ok(())
     }


Lethe10137:
> @Lethe10137, is it still a draft PR or ready for review?

I believe it is ready and have sent request for review.

rattan-core/src/cells/per_packet/delay.rs
```diff
@@ -147,13 +146,16 @@ where
...
         if instant <= Instant::now() {
+            packet.delay_until(instant);
             Some(packet)
         } else {
             // We shouldn't have dequeue the packet, it is not ready to be sent yet.
-            self.packet_queue.renqueue(packet, instant);
-            None
+            // self.packet_queue.renqueue(packet, instant);
+            // None
+            todo!()
```
BobAnkh: Why you change this to `todo()!`? This seems to be not complete. Or why we can't just return None and reenqueue? Are you assuming this branch will never trigger, otherwise, this will panic at runtime.

rattan-core/src/cells/mod.rs
```diff
@@ -78,10 +120,10 @@ impl Packet for StdPacket {
         }
     }
 
-    fn from_raw_buffer(buf: &[u8]) -> Self {
+    fn from_raw_buffer(buf: &[u8], timestamp: Instant) -> Self {
```
BobAnkh: I wonder if this is necessary, as I find that most of your real use cases just use Instant::now when calling this method. This is really a breaking change where I think is not necessary. I think we won't set this to another timestamp other than Instant::now in real usage, right? For test cases, you can just directly change this field's value.

rattan-core/src/cells/mod.rs
```diff
@@ -362,3 +501,91 @@ macro_rules! wait_until_started {
...
+pub enum State {
+    // Drops all packets
+    Drop,
+    // Passes through all packets
+    PassThrough,
+    // Normal operation
+    Normal,
+}
+
+pub struct AtomicState(AtomicU8);
```
BobAnkh: Why is this only used in the delay cell, not in other cells? This is not consistent.

Also, since you defined `State`, I think we should modify the `change_state` function signature to use `State`? Also, rename `State` to `CellState` to make it more understandable

rattan-core/src/cells/delay.rs
```diff
@@ -52,25 +49,75 @@ where
...
+    //             // then the packet leaves immediately and no further new delay apply
+    //             packet_delay = *timestamp - packet_timestamp;
+    //             break;
+    //         } else {
+    //             packet_delay = *delay
+    //         }
+    //     }
+    //     packet_delay
+    // }
+    // ************ Which behavior here is right is under discussion. ***********
```
BobAnkh: Shall we remove these comments as they seems not necessary (you can just remove it once I approved this PR. You can keep it for reference before that)

rattan-core/src/cells/delay.rs
```diff
@@ -443,13 +496,13 @@ where
...
-                current_delay: Delay::ZERO,
-                next_available: Instant::now(),
+                delays: Configs::new(),
                 next_change: Instant::now(),
                 config_rx,
                 send_timer: Timer::new()?,
                 change_timer: Timer::new()?,
-                state: AtomicI32::new(0),
+                state: AtomicState::new(State::Drop),
+                latest_egress_timestamp: Instant::now() - Duration::from_secs(10),
```
BobAnkh: The same as above, this maybe a potential bug.

BobAnkh: And as we use the 10 many times, we should define the Duration::new(10,0) as a const var.

rattan-core/src/cells/delay.rs
```diff
@@ -52,25 +49,75 @@ where
...
+    //             packet_delay = *delay
+    //         }
+    //     }
+    //     packet_delay
+    // }
+    // ************ Which behavior here is right is under discussion. ***********
+    fn delay(&mut self, packet_timestamp: Instant) -> Delay {
+        // Eliminate out-dated items from the Configs<Delay>
+        // The correctness of this function relies on that the `pakcet_timestamp` never decreases.
+
```
BobAnkh: I think we shall not use this kind of get delay, which is not logically fluent. Please refer to my other comments that we shall make configuration simpler.

rattan-core/src/cells/mod.rs
```diff
@@ -362,3 +501,91 @@ macro_rules! wait_until_started {
...
+            expire: None,
+        }
+    }
+    fn insert(&mut self, time: Instant, data: Option<C>) {
+        if let Some(data) = data {
+            self.configs.insert(time, data);
+            self.expire = None;
+        } else {
+            self.expire = Some(time);
+        }
```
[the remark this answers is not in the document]
BobAnkh: Besides, you set expire, but where do you use it?

rattan-core/src/metal/timer.rs
```diff
@@ -55,4 +59,12 @@ impl Timer {
...
         }
     }
+
+    pub async fn sleep_until(&mut self, instant: Instant) -> Result<Instant, MetalError> {
+        match instant.duration_since(Instant::now()) {
+            Duration::ZERO => tokio::task::yield_now().await,
+            sleep_time => self.sleep(sleep_time).await?,
+        }
+        Ok(instant)
+    }
```
BobAnkh: I think we could make this more explicit:

```rust
pub async fn sleep_until(&mut self, instant: Instant) -> Result<Instant, MetalError> {
    let now = Instant::now();
    if instant <= now {
        tokio::task::yield_now().await;
    } else {
        self.sleep(instant - now).await?;
    }
    Ok(instant)
}

```

rattan-core/src/cells/delay.rs
```diff
@@ -198,12 +256,13 @@ where
...
                 egress: tx,
-                delay,
+                delays: Configs::from([(Instant::now(), delay)]),
                 config_rx,
                 timer: Timer::new()?,
-                state: AtomicI32::new(0),
+                state: AtomicState::new(State::Drop),
                 notify_rx: None,
                 started: false,
+                latest_egress_timestamp: Instant::now() - Duration::from_secs(10),
```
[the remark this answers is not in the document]
BobAnkh: Can we just init it to the start time of rattan?

BobAnkh (changes requested):
I have finished a basic review on the changes except for token bucket. I think we should first resolve these comments and then I will continue to review for token bucket in case there are any further changes.

rattan-core/src/cells/delay.rs
```diff
@@ -569,12 +626,17 @@ mod tests {
         let received = received.unwrap();
         assert!(received.length() == 256);
 
-        assert!((duration - 20.0).abs() <= DELAY_ACCURACY_TOLERANCE);
+        // assert_eq!(received.delay(), Duration::from_millis(20));
+        // assert!((duration - 20.0).abs() <= DELAY_ACCURACY_TOLERANCE);
+        // ************ Which behavior here is right is under discussion. ***********
+        // For simplicity, only consider how long should a packet be delayed based on the latest config upon the packet's timestamp
```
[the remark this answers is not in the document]
Lethe10137: Summary: I redesigned "which config should be applied" to a much simpler version. 

Basic idea is, by warping the trace's interface as a peekable iterator, to maintain:
1)  A "current" config, which is expected to be used in most cases, with a explicit valid period (not valid before & not valid after)
2)  Last "current" config,  which is automatically maintained as last "current" config set. Apply for any packet that timestamped prior to the valid period of the current config. Also used as a fallback value. 
3)  A "next" config which is peeked from the iterator. Applied for packets that cames after the expiry of current config, yet before the config was updated (which should have happened but was delayed).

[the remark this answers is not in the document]
BobAnkh: As I commented below, I'm not sure if "peekable" is necessary. So why would happened of a future-packet handled with config not updated?

Lethe10137:
> @Lethe10137 I think the overall structure of this new design is good, but I am still not particularly sure about some details. On one hand, I don't quite understand the significance of the next field; could you explain it more or clarify whether it would be problematic or difficult to write code without this field? On the other hand, I'm also unclear about when we will update this IncomingConfig. Will there be situations where we need to use next? In what circumstances would that apply?

For example in the test of `test_replay_cell_change_state`, when no packet keep the egress busy.
```
2025-11-07T04:48:01.867022Z DEBUG test_replay_delay_cell_change_state: rattan_core::cells::incoming_config: get_current called at logical 2.102035856s wallclock 2.102065406s
2025-11-07T04:48:01.867087Z DEBUG test_replay_delay_cell_change_state: rattan_core::cells::delay: Setting 50ms delay valid from 1s till 2s
2025-11-07T04:48:01.867107Z DEBUG test_replay_delay_cell_change_state: rattan_core::cells::incoming_config: update called at logical 1s wallclock 2.10215119s
2025-11-07T04:48:01.867118Z DEBUG test_replay_delay_cell_change_state: rattan_core::cells::incoming_config: get_current called at logical 2.102035856s wallclock 2.102161953s
2025-11-07T04:48:01.867164Z DEBUG test_replay_delay_cell_change_state: rattan_core::cells::delay: Setting 10ms delay valid from 2s till 3s
2025-11-07T04:48:01.867179Z DEBUG test_replay_delay_cell_change_state: rattan_core::cells::incoming_config: update called at logical 2s wallclock 2.102223425s
```
The logical time of config changes, based on the `next_change`, could significantly lack behind wallclock when there was no packets to keep the egress busy.  

I rethought about it and perhaps there's a way out without a `next` field. 
In the egress for a replay cell , use a while loop to make sure that  packet.timestamp < self.next_change  before entering the `tokio::select`.

Imagine that we have a trace that changes every 1ms, and It was not before 10s since the trace begin(the first packet) until the second packet entered rattan.  Upon processing the second packet, we have to call `self.update_delay` 10,000 times as long as the trace is an iterator, and it would be quicker to called in a simple loop rather than in the loop of `tokio::select!`, perhaps?

PR #150, opened 2025-11-26
Pktlog converter & combined-trace

Lethe10137:
1.  A new sub command `rattan convert` to convert Rattan packet log into `.pcapng` file. This also comes with a new design of Rattan packet log, which is implemented in a the `rattan-log` crate.
2.  A new option `--combiend-trace`, allowing export the traces of Bandwidth, Delay and Loss Cells together in a csv file (it works well with Libra Office Calc).

### DIFF
--- a/guide/src/ref/packet-log.md
+++ b/guide/src/ref/packet-log.md
@@ -1,37 +1,231 @@
 # Packet Log
 
-The packet log contains a binary file that contains a compressed record of all packets processed by Rattan and a flow description file. It is designed to store what happened during the experiment with low overhead, enabling post-analysis and debugging.
+The Rattan packet log contains the following binary files:
+
+- A `.flow` file, which contains a basic description of all (recognized) flows processed by Rattan.
+- A `.rtl` file, which contains a compressed record of the packets in those flows.
+- Optionally, a `.raw` file, which contains the protocol headers of those packets.
+
+It is designed to store what happened during the experiment with low overhead, enabling post-analysis and debugging.
+
+**Notice**:
+Rattan currently records TCP flows on IPv4 only. Any other packets that are not recognized as a TCP flow on IPv4 are ignored
+when recording a Rattan packet log, while they are normally processed and forwarded by
+Rattan (e.g., UDP flows).
 
 ## Usage
 
-Using the `--packet-log` option in the CLI will enable the packet log feature and specify the file to store the records. The flow description file will be stored in the same directory with the name `<packet_log>.flows`, where `<packet_log>` is the name of the packet log file.
+### Record a Rattan packet Log
 
-For example, you can run the following command to enable packet logging:
-
-```bash
-rattan link --packet-log packet.log
-```
-
-This will create a file named `packet.log` in the current directory, which will contain the compressed packet log. The flow description file will be named `packet.log.flows`.
-
-We also provide a python script [scripts/log_converter.py](https://github.com/stack-rs/rattan/blob/main/scripts/log_converter.py) to convert the packet log file (along with the flow description file) to some pcapng files, which can be viewed with Wireshark or other tools. You can run the script like this:
+Two options are involved: `--packet-log` and `--packet-log-mode` . Example usages are:
 
 ```bash
-python3 scripts/log_converter.py packet.log converted.pcapng
+rattan link --packet-log packet.rtl --packet-log-mode compact-tcp
+rattan run --config config.toml --packet-log packet.rtl --packet-log-mode raw-ip
 ```
 
-This will read `packet.log` and `packet.log.flows`, and generate a pcapng file for each end of Rattan (identified by IP address), with the name `converted_<ip>.pcapng` (e.g. `converted_10_1_1_1.pcapng`).
+The possible values for `--packet-log-mode` are:
 
-Note that some fields are artificially constructed in the pcapng file, as the compressed packet log only stores partial information. Artificial fields include:
+- `compact-tcp` (as default, if not specified). In this mode, some fields from TCP and IP headers are recorded.
+- `raw-ip`. In this mode, the raw L3 and L4 headers are recorded.
+- `raw-tcp`. In this mode, the raw L4 headers are recorded.
+
+This will create a file named `packet.rtl` in the current directory, which will contain the compressed packet log.
+The flow description file will be named `packet.flows`. Also, if any raw headers are recorded, a `packet.raw` will be created.
+
+### Convert a Rattan packet log
+
+#### Rust subcommand
+
+To convert a Rattan packet log to a `.pcapng` file, which can be viewed with Wireshark or other tools, we provide
+a subcommand `convert`.
+
+Example usage:
+
+```bash
+rattan convert ./packet.rtl [path/to/output]
+```
+
+This expects the following files exist, as input:
+
+- `./packet.rtl`
+- `./packet.flows`
+- `./packet.raw` (optional, only if the packet log was recorded in `raw-tcp` or `raw-ip` mode).
+
+And it will generate the following files, as output, if the address of the two ends of the emulated link
+of Rattan was `10.1.1.1` and `10.2.1.1`":
+
+- `path/to/output_10_1_1_1.pcapng`
+- `path/to/output_10_2_1_1.pcapng`
+
+If the second argument was not provided (the output prefix), the default value for it is the first argument (the path
+to the `.rtl` file).
+
+Note that, in `compact-tcp` mode, some fields are artificially constructed in the pcapng file, as the compressed
+packet log only stores partial information. Artificial fields include:
 
 - `Window`: The TCP window size is set to `65535` for all packets.
 - `Checksum`: The TCP checksum is set to `0` for all packets.
 - `Urgent Pointer`: The TCP urgent pointer is set to `0` for all packets.
 - `Options`: The TCP header length is real, but option content is set to `0` for all packets.
 
+Besides, in `raw-tcp` mode, some fields are artificially constructed in the pcapng file, since per-packet
+info in IP headers are not recorded at all. Artificial fields include:
+
+- `Identification`: Set to `0`.
+- `Flags` : Set to `Don't Fragment`.
+- `Checksum` : Set to `0`, or disabled.
+
+#### Python script
+
+We provide a python script ([scripts/log_converter.py](https://github.com/stack-rs/rattan/blob/main/scripts/log_converter.py)), that provides basically the same function as the above-mentioned Rust subcommand does.
+Notice that the Python script can be slower than the Rust subcommand by one to
+two orders of magnitude when processing large log files.
+
+Example usage:
+
+```bash
+python scripts/log_converter.py ./packet.rtl path/to/output
+```
+
+or, use the script as a [uv script](https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file) to automatically handle dependencies.
+
+```bash
+uv run --script scripts/log_converter.py ./packet.rtl path/to/output
+```
+
+The paths to input and output files are the same as those for `rattan convert`, and the behaviour of the python script is basically the same as `rattan convert`.
+
 ## Specification
 
-The packet log is stored in system endianness (little-endian in most cases and in our tests), entry by entry. The specification for each log entry is as follows:
+### Raw file
+
+A `.raw` file is an unstructured binary file. Currently, the headers recorded
+in `raw-tcp` and `raw-ip` modes are directly written into the `.raw` file continuously,
+without any padding, header or index.
+The file is considered a byte buffer, where a slice (starting from `offset` bytes from
+the head of the file and ending at `offset` + `length`)
+can be indexed by the pair (`offset`, `length`).
+
+### Log entry file
+
+Now we have 2 kinds of log entry files:
+
+- `.flows` files, which contains the metadata for flows. They consist of `Flow Entries`.
+- `.rtl` files, which contains per-packet log for packets in those flows.
+  - For non-raw modes, they consist of non-raw `General Packet Entries`.
+  - For raw modes, they consist of chunks. Each chunk starts with a `Chunk Prologue`, which is followed by raw `General Packet Entries`.
+
+Above mentioned log entry types are described as follows.
+
+Every log entry file consists of `Log Entries`.
+Those log entries are stored in system endianness (little-endian in most cases and in our tests).
+The specification for a log entry is as follows:
+
+```text
+0                   1                   2                   3
+0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|       LH.length       | LH.ty.|                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+                               +
+|                                                               |
++                                                               +
+|                            payload                            |
++                                                               +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
+
+```bash
+protocol "LH.length:12,LH.ty.:4,payload:112"
+```
+
+Generated with [protocol](https://github.com/luismartingarcia/protocol), where
+`LH` is short for log entry header.
+
+`LH.length` denotes the whole length of the log entry file, including the header (2 bytes) and its payload, in bytes.
+
+Currently, there are multiple types of `LH.ty.` that indicate how a log entry's payload is organized. They are:
+
+- `LH.ty.` = 0, for a General Packet Entry as payload.
+- `LH.ty.` = 1, reserved for future use.
+- `LH.ty.` = 2, for a Flow Entry as payload,
+- `LH.ty.` = 15, for a Chunk Prologue as payload,
+
+#### Chunk Prologue Entry
+
+A chunk prologue entry is a log entry with `LH.ty.` = 15, which is used in raw-mode packet logs.
+
+```text
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|       LH.length       | LH.ty.|       CPH.length      |CPH.ty.|
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                        CPH.log_version                        |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                        CPH.data_length                        |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                        CPH.chunk_length                       |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                                                               +
+|                                                               |
++                           CP (custom)                         +
+|                                                               |
++                                                               +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
+
+```bash
+protocol "LH.length:12,LH.ty.:4,CPH.length:12,CPH.ty.:4,CPH.log_version:32,CPH.data_length:32,CPH.chunk_length:32,CP (custom):128"
+```
+
+Where `CPH` is short for Chunk Prologue Header, and `CP` is short for Chunk Prologue.
+The meaning of each field is as follows:
+
+- `CPH.length` denotes the length of the chunk prologue entry itself.
+- `CPH.ty.` is always 0 now.
+- `CPH.log_version` is a magic number, which is 0x20260101 in this specification.
+- `CPH.data_length` is the sum of the length of this Chunk Prologue Entry and all subsequent content (which, for now, consists solely of raw General Packet Entries).
+- `CPH.chunk_length` is the logical length of the current chunk. It indicates that the next Chunk Prologue is expected to be found, or the End-of-File (EOF) is expected to be reached, exactly `CPH.chunk_length` bytes from the offset of this log entry.
+
+And when `CPH.ty.` is 0, the body part of the Chunk Prologue is:
+
+```text
+ 0                   1                   2                   3
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                           CP.offset                           +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                                                               |
++                            Reserved                           +
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
+
+```bash
+protocol "CP.offset:64,Reserved:64"
+```
+
+Where `CP` is short for Chunk Prologue.
+`CP.offset` is a 64-bit integer, whose value is referenced by the raw General Packet entries in the chunk that this
+Chunk Prologue introduces.
+
+#### General Packet Entry
+
+A General Packet Entry is a log entry with `LH.ty.` = 0, which records a packet.
+
+There are two modes of General Packet Entries:
+
+- non-raw (`GPH.ty.` = 0), and
+- raw (`GPH.ty.` = 1, 2).
+
+**non-raw General Packet Entry**:
+The specification for a non-raw General Packet Entry:
 
 ```text
  0                   1                   2                   3
@@ -47,15 +241,12 @@ The packet log is stored in system endianness (little-endian in most cases and i
 +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ```
 
-Generated with [protocol](https://github.com/luismartingarcia/protocol), where:
-
-- `LH` is short for log entry header
-- `GPH` is short for general packet entry header
-- `GP` is short for general packet entry (the body part)
-- `PRH` is short for protocol entry header
-- `PRT` is short for protocol entry (the body part)
-- `ty.` is short for type
-- `ac.` is short for action
+- `GPH` is short for general packet entry header.
+- `GP` is short for general packet entry (the body part).
+- `PRH` is short for protocol entry header.
+- `PRT` is short for protocol entry (the body part).
+- `ty.` is short for type.
+- `ac.` is short for action.
 
 ```bash
 protocol "LH.length:12,LH.ty.:4,GPH.length:8,GPH.ac.:4,GPH.ty.:4,GP.timestamp:32,GP.length:16,PRH.length:12,PRH.ty.:4,PRT (custom):32"
@@ -64,14 +255,14 @@ protocol "LH.length:12,LH.ty.:4,GPH.length:8,GPH.ac.:4,GPH.ty.:4,GP.timestamp:32
 The meaning of each field is as follows:
 
 - `LH.length`: Length of the whole log entry in bytes. For TCP log entries, it is 32 bytes.
-- `LH.ty.`: Type of the log entry, currently only `0` is supported (indicating a general packet log entry).
+- `LH.ty.`: Type of the log entry, currently only `0` is supported (indicating a general packet entry).
 - `GPH.length`: Length of the general packet entry in bytes, including the header and body (`GPH` + `GP`). It is 8 bytes in most cases.
 - `GPH.ac.`: Action of the general packet entry. It can be one of the following values:
   - `0`: Send (packet leaves Rattan)
   - `1`: Receive (packet enters Rattan)
   - `2`: Drop (packet is dropped by Rattan)
   - `3`: Passthrough (packet passes through Rattan without any effect applied)
-- `GPH.ty.`: Type of the general packet entry, currently only `0`.
+- `GPH.ty.`: Type of the general packet entry, `0` for non-raw GPH.
 - `GP.timestamp`: Timestamp of the packet entry in microseconds since `base_ts` (in flow description file).
 - `GP.length`: Length of the whole packet in bytes.
 - `PRH.length`: Length of the protocol entry in bytes, including the header and body (`PRH` + `PRT`). For TCP log entries, it is 22 bytes.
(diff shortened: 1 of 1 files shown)


guide/src/guide/log.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Raw
+
+In such modes, the raw headers (L4 for `RawTCP` and L3 + L4 for `RawIP`) of the packets of the designated protocol is recorded, and Rattan generates 3 log files during runtime.
+
+1. `xxxxx.flow`.  Same as above.
+2. `xxxxx.raw`. A un-structed file, that contains the raw, unaligned packet headers continously.
+3. `xxxxx.rtl`. which contains the `Raw Packet Log Entries`. The different between these entries and the above describled counterparts (those in `oldrtl`) is that, they use a "fat pointer" (i.e., start offset and size) to the `.raw` file, pointing at the raw recorded header of the packet that is related to the entry.
+
+The `rattan convert` subcommand is also appliable for raw entries. Given that more information of packets are recorded, it is expected that the recovered `.pcapng` file from raw logs
```
BobAnkh: Move to a new section as I mentioned above.

guide/src/guide/log.md
```diff
@@ -0,0 +1,43 @@
...
+1. `xxxxx.flow`, which contains flow level metadata info.
+2. `xxxxx.oldrtl`, which contains the `Packet Log Entries`.  Each entry here contains some fields from header, along with some packet-level meta data (e.g. timestamp, length) and is related to a certain flow recorded as above.
+
+We provide `rattan convert` subcommand to use above log info to (partially) recover a `.pcapng` file that can be read by standard tools such as wireshark.
+
+### Raw
+
+In such modes, the raw headers (L4 for `RawTCP` and L3 + L4 for `RawIP`) of the packets of the designated protocol is recorded, and Rattan generates 3 log files during runtime.
+
+1. `xxxxx.flow`.  Same as above.
```
BobAnkh: You use two spaces here. Please do format your markdown file.

guide/src/guide/log.md
```diff
@@ -0,0 +1,43 @@
...
+2. `xxxxx.raw`. A un-structed file, that contains the raw, unaligned packet headers continously.
+3. `xxxxx.rtl`. which contains the `Raw Packet Log Entries`. The different between these entries and the above describled counterparts (those in `oldrtl`) is that, they use a "fat pointer" (i.e., start offset and size) to the `.raw` file, pointing at the raw recorded header of the packet that is related to the entry.
```
BobAnkh: "they use a fat pointer" -> "they contains a fat pointer targeting a part of the `.raw` file"

rattan-core/src/cells/external.rs
```diff
@@ -26,7 +28,8 @@ use serde::Deserialize;
 use tokio::{io::unix::AsyncFd, sync::mpsc::UnboundedSender};
 use tracing::{debug, error, instrument};
 
-use super::{ControlInterface, Egress, FlowDesc, Ingress};
+use super::{ControlInterface, Egress, Ingress};
+use rattan_log::log_entry::PktAction;
```
BobAnkh: Why is this a separate import sentence? Please merge it into the above.

scripts/log_converter.py
BobAnkh: While this is not necessary, we could we also preserve a python version that can read, parse and convert our log as an example and a inter-verification testing.

Cargo.toml
```diff
@@ -8,57 +8,62 @@ repository.workspace = true
...
 ] }
 tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
-
-[features]
-default = ["first-packet"]
-first-packet = ["rattan-core/first-packet"]
-http = ["rattan-core/http"]
-packet-dump = ["rattan-core/packet-dump"]
+test-log = { version = "0.2.18", features = ["trace"] }
```
BobAnkh: This is not a dependency? This is a dev-dependency ?

Cargo.toml
```diff
@@ -8,57 +8,62 @@ repository.workspace = true
...
 
 [dependencies]
+clap = { version = "4.2.4", features = ["derive"] }
```
BobAnkh: Please use `clap = { workspace = true }`

src/combined_trace.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+// Define output format of an delay item in the .csv file here
+fn express_delay(delay: Duration) -> String {
+    format!("{:?}", delay.as_secs_f64())
+}
+
+// Define output format of the trace logical timestamp in the .csv file here
+fn express_time(time: Duration) -> String {
+    format!("{:?}", time.as_secs_f64())
+}
```
BobAnkh: We should also support human-readable format in our output.

src/main.rs
```diff
@@ -56,6 +67,20 @@ pub struct Arguments {
...
+    #[arg(
+        long,
+        global = true,
+        value_parser = parse_duration
+    )]
+    combined_trace: Option<Duration>,
+
+    /// Generate combined trace csv file to the specified path instead of stdout
+    #[arg(long, requires = "combined_trace", global = true, value_name = "File")]
+    combined_trace_path: Option<PathBuf>,
```
BobAnkh: We should let user specify the output format, start_time and end_time. For the output path, we should change the extension according to the format.

src/main.rs
```diff
@@ -120,6 +149,8 @@ enum CliCommand {
     Link(channel::ChannelArgs),
     /// Run the instance according to the config.
     Run(RunArgs),
+    /// Convert Rattan packet log into .pcapng file.
+    Convert(LogConverterArgs),
```
BobAnkh: Change the name of the struct to `ConvertLogArgs`

rattan-log/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+use std::path::PathBuf;
+
+/// Convert Rattan Packet Log file to pcapng file for each end
+#[derive(Args, Debug, Default, Clone)]
+#[command(rename_all = "kebab-case")]
+pub struct LogConverterArgs {
+    /// Input Rattan Packet Log file path
+    pub input: PathBuf,
+    /// Output pcapng file name prefix. take the input as default value
+    pub output: Option<PathBuf>,
```
BobAnkh: This shall not be in here (not in this crate)

rattan-log/src/logger/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl Eq for FlowDesc {}
+
+#[derive(Clone)]
+pub enum RattanLogOp {
+    Entry(Vec<u8>),
+    RawEntry(u32, RawLogEntry, Vec<u8>), // Entry, Raw header
+    Flow(u32, i64, FlowDesc),
+    End,
```
BobAnkh: Should give this enum some comments on each field and each value. What do they mean?

rattan-log/src/logger/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (FlowDesc::TCP(a0, b0, c0, d0, _), FlowDesc::TCP(a1, b1, c1, d1, _)) => {
+                (a0, b0, c0, d0) == (a1, b1, c1, d1)
+            }
+        }
+    }
+}
+
+impl Eq for FlowDesc {}
```
BobAnkh: Directly derive this instead of write a impl block if possible.

rattan-log/src/logger/build_pcap.rs
```diff
@@ -0,0 +1,190 @@
+use crate::{log_entry::entry::flow_entry::TCPFlow, RawLogEntry, TCPLogEntry};
```
BobAnkh: For these types, we should expose them in the root of this crate.

rattan-log/src/logger/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+use tokio::sync::mpsc::UnboundedSender;
+
+use std::hash::Hash;
+use std::net::Ipv4Addr;
+
+use crate::RawLogEntry;
+
+pub mod build_pcap;
+pub mod file_reader;
+pub mod file_writer;
```
BobAnkh: Let's just rename them to `pcap`, `reader` and `writer`.

rattan-log/src/logger/mmap.rs
```diff
@@ -0,0 +1,485 @@
...
+    fn new(file: &File, offset: u64, length: usize) -> Result<Self> {
+        Ok(Self {
+            mmap: mmap_mut_segment(file, offset, length)?,
+            length,
+            written: 0,
+        })
+    }
+}
+
+pub struct MmapWritter<const P: usize> {
```
BobAnkh: Should be MmapWriter, spell typo.

rattan-log/src/logger/mmap.rs
```diff
@@ -0,0 +1,485 @@
...
+
+    use random::Source;
+    use tempfile::tempdir;
+
+    fn new_file(name: &str) -> Result<PathBuf> {
+        let dir = tempdir()?;
+        Ok(dir.path().join(name))
+    }
+
+    fn stream_writting(pattern: impl Iterator<Item = usize>) {
```
BobAnkh: Should be `stream_writing`, typo.

rattan-log/src/logger/mmap.rs
```diff
@@ -0,0 +1,485 @@
...
+    path: PathBuf,
+    chunk_start: u64,
+    chunk: MmapChunk,
+}
+
+impl<const P: usize> Drop for MmapWritter<P> {
+    fn drop(&mut self) {
+        let file_last_write = self.offset();
+        let _ = self.file.set_len(file_last_write);
+        tracing::info!(
```
BobAnkh: This shall not be info, but debug.

rattan-log/src/logger/mmap.rs
```diff
@@ -0,0 +1,485 @@
...
+    init: bool,
+}
+
+impl<const P: usize, H, C> Drop for MmapChunkWriter<P, H, C>
+where
+    H: FnMut(usize, &Option<C>) -> Vec<u8>,
+{
+    fn drop(&mut self) {
+        self.finish_chunk(true);
+        tracing::info!("Mmap total size: {:?}", self.writer.offset());
```
BobAnkh: This shall not be info but debug.

src/combined_trace.rs
```diff
@@ -0,0 +1,170 @@
...
+const DELAY_TRACE_SUFFIX: &str = "_delay_secs";
+
+pub fn write_combined_trace<P: Packet>(
+    output: impl Write,
+    cells: HashMap<String, CellBuildConfig<P>>,
+    length: Duration,
+) -> Result<()> {
+    let mut cells_to_report = vec![];
+
+    let mut change_points: BTreeMap<Duration, Vec<(usize, String)>> = BTreeMap::new();
```
BobAnkh: For very long traces or high-frequency changes, this might consume a significant amount of memory. If this becomes an issue, a streaming approach (using a k-way merge of iterators from the traces) could be considered. I don't know if it is easier to implement a more streaming approach. If so, we can implement it, if not, we should comment it here and we can leave it for future.

rattan-log/src/logger/file_reader.rs
```diff
@@ -0,0 +1,260 @@
...
+    let flow_file: Mmap = load_raw_file(&flow_file_path)?;
+
+    let mut raw_file_path = file_path.as_ref().to_path_buf();
+    raw_file_path.set_extension("raw");
+
+    // Unused if the log is not in raw format.
+    let mut raw_file: Option<Mmap> = None;
+    fn load_raw_file(path: &PathBuf) -> Result<Mmap> {
+        mmap_file(&File::open(path)?)
+    }
```
BobAnkh: Move the definition up to just below the outer funciton signature (line 75)

rattan-log/src/logger/file_writer.rs
BobAnkh: Please check and make sure your writer and reader can be interrupted and released normally ( when we Ctrl-C while running Rattan). I don't check its logic in such depth, so you please check it for me and see if its loop (for both reader and writer) can successfully break and release the resource upon SIGTERM.

rattan-log/src/log_entry/entry/tcp_ip_compact.rs
```diff
@@ -0,0 +1,72 @@
...
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+#[derive(Debug, Clone, Copy, Default)]
+#[repr(C, packed(2))]
+pub struct TCPLogEntry {
+    pub header: LogEntryHeader,
+    pub general_pkt_entry: GeneralPktEntry,
+    pub tcp_entry: TCPProtocolEntry,
+}
+unsafe impl Plain for TCPLogEntry {}
+impl TCPLogEntry {
```
BobAnkh: Some new lines are needed between them.

rattan-log/src/log_entry/entry/flow_entry.rs
```diff
@@ -0,0 +1,208 @@
...
+// |                                                               |
+// |                                                               |
+// |                                                               |
+// |              Options on SYN/SYN_ACK (40Bytes)                 |
+// |                                                               |
+// |                                                               |
+// |                                                               |
+// |                                                               |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+//
```
BobAnkh: This should be directly attached to the time saying if we have such type (I'm not sure). If we do not have such a type containing these all, then this protocal definition can just leave here, otherwise, move it to that type.

rattan-log/src/log_entry/entry/chunk_header.rs
```diff
@@ -0,0 +1,93 @@
...
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                   offset (lower 32bits)                       |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                   offset (upper 32bits)                       |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                          reserved                             |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                          reserved                             |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
```
BobAnkh: Please remove this empty line

rattan-log/src/log_entry/entry/chunk_header.rs
```diff
@@ -0,0 +1,93 @@
...
+// |                        chunk  length                          |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                   offset (lower 32bits)                       |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                   offset (upper 32bits)                       |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                          reserved                             |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                          reserved                             |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
BobAnkh: Please use the tool protocol to generate all such definitions again across the crate. Your filed name should not have two or even more spaces between the words!!

rattan-core/src/cells/mod.rs
```diff
@@ -1,12 +1,11 @@
...
 use etherparse::{Ethernet2Header, Ipv4Header};
+use rattan_log::FlowDesc;
 #[cfg(feature = "serde")]
-use serde::{Deserialize, Serialize};
-use std::{fmt::Debug, net::Ipv4Addr, sync::Arc};
+use serde::Deserialize;
+use std::{fmt::Debug, sync::Arc};
 use tokio::time::Instant;
-
-use crate::error::Error;
```
BobAnkh: Please correct the import order according to my guidance in other comments.

rattan-core/src/config/mod.rs
```diff
@@ -8,6 +8,8 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "http")]
 use crate::control::http::HttpConfig;
 
+use crate::radix::PacketLogMode;
```
BobAnkh: Please let's fix the import orders.

rattan-core/src/radix/mod.rs
```diff
@@ -33,14 +32,30 @@ use crate::{control::http::HttpControlEndpoint, error::HttpServerError};
...
+    RawTCP,
+}
+
+#[cfg(not(feature = "serde"))]
+#[derive(Clone, Copy, Debug, clap::ValueEnum, PartialEq, Eq)]
+pub enum PacketLogMode {
+    CompactTCP,
+    RawIP,
+    RawTCP,
+}
```
BobAnkh: Do not define the struct twice. You should use `#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]` instead.

BobAnkh (changes requested):
Let's replace `unwrap` with `expect` as much as possible.

And please check all the import order and the positions of your imports and consts/static vars.

Check all the places where you use tracing, and ensure that all instances of tracing::info are indeed necessary for outputting information that users need to see.

Another important thing to remember is that: your typo or style should be compatible across all your changes! Please leave new lines between code blocks, code up the comments, etc.

src/combined_trace.rs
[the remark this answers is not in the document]
Lethe10137: 3 Human Readable versions:
CSV TSV JsonHumanReadable are supported
and `Json` is supported to be read by machine:
```json
[{"down_1_loss_pattern":[0.0],"duration_secs":{"nanos":0,"secs":1},"start_time_secs":{"nanos":0,"secs":0},"up_1_bandwidth":{"bps":200000000,"gbps":1},"up_2_delay":{"nanos":10000000,"secs":0}},{"down_1_loss_pattern":[0.5,0.7,0.2],"duration_secs":{"nanos":341000000,"secs":0},"start_time_secs":{"nanos":0,"secs":1},"up_1_bandwidth":{"bps":200000000,"gbps":1},"up_2_delay":{"nanos":10000000,"secs":0}}]

```

rattan-core/src/cells/mod.rs
Lethe10137: I suppose that the order is:
1. super
2. crate
3. other crates in this workspace
4. std
5. 3rd-party crates

Is this correct?

rattan-log/src/log_entry/entry/flow_entry.rs
Lethe10137: We do have a type that contains these all, and it is `TCPFlowEntry`, placed 4 lines after this

src/combined_trace.rs
Lethe10137: I commented it

BobAnkh:
> @BobAnkh while the CI is still broken, could you please have another review of this PR? I tried to handle each of the conversations above, and have checked and fixed some editorial issues.

Ok, I will review it. We can just ignore the broken CI. I will review it. If all things are good, please solve the conflicts. Do not rebase or force-push the branch while we are still reviewing and modifying.

guide/src/ref/packet-log.md
```diff
@@ -1,37 +1,167 @@
...
-Using the `--packet-log` option in the CLI will enable the packet log feature and specify the file to store the records. The flow description file will be stored in the same directory with the name `<packet_log>.flows`, where `<packet_log>` is the name of the packet log file.
+### Record a Rattan packet Log
 
-For example, you can run the following command to enable packet logging:
+Two options are involved: `--packet-log` and `--packet-log-mode` . Example usages are:
 
 ```bash
-rattan link --packet-log packet.log
+rattan link --packet-log packet.rtl --packet-log-mode compact-tcp
+rattan run --config CONFIG_FILE --packet-log packet.rtl --packet-log-mode raw-ip
```
BobAnkh: Should be `<CONFIG_FILE>` not `CONFIG_FILE` if you use this to indicate that user should use their config file names, or you could just use `config.toml`.

guide/src/ref/packet-log.md
```diff
@@ -1,37 +1,167 @@
...
+- `./packet.flows`
+- `./packet.raw` (optional, only if the packet log was recorded in `raw-tcp` or `raw-ip` mode).
+
+And it will Generated the following files, as output, if the address of the two ends of the emulated link
+of Rattan was `10.1.1.1` and `10.2.1.1`":
+
+- `./output_10_1_1_1.pcapng`
+- `./output_10_2_1_1.pcapng`
+
+If the second argumnet was not provided(the output prefix), the default value for it is the first argument (the path
```
BobAnkh: Also, I wonder if it is possible that I specify a directory-like prefix. Suppose I run:

```bash
rattan convert ./packet.rtl ./output/layer3
```

What will happen? Will it produce files to ./output/layer3_10_1_1_1.pcapng ? Whether this is correct or not, we should note this in our documentationi.

scripts/log_converter.py
BobAnkh: As this is a script, we shall not provide such a complete python project structure. So I wonder could you please try the uv script structure: [https://docs.astral.sh/uv/guides/scripts/](https://docs.astral.sh/uv/guides/scripts/). We can have just the log_converter.py file without any other files like pyproject.toml or uv.lock

guide/src/ref/packet-log.md
```diff
@@ -1,37 +1,167 @@
...
+in it can be indexed by the pair (`offset`, `length`).
+
+### Log entry file
+
+Now we have 2 kinds of log entry files:
+
+- `.flows` files, which contains the metadata for flows. They consist of `Flow Entries`.
+- `.rtl` files, which contains per-packet log for packets in those flows.
+  - For non-raw modes, they consist of non-raw `General Packet Entries`.
+  - For now modes, they consist of chunks. Each chunk starts with a `Chunk Prologue`, which is followed by raw `General Packet Entries`.
```
BobAnkh: What is `For now modes`? I assume it should be `For raw modes`

rattan-log/src/logger/reader.rs
```diff
@@ -0,0 +1,260 @@
...
+            }
+        }
+
+        let Ok(log_entry) = LogEntry::read(&mut log_entries) else {
+            if current_chunk_end.is_none() {
+                // If there is no chunk header
+                break;
+            }
+            continue;
+        };
```
BobAnkh: If `LogEntry::read` fails (e.g., due to a corrupted entry or unexpected EOF within a chunk) without advancing the cursor, and `current_position < current_chunk_end`, the loop might spin infinitely because the `continue` jumps back to the start, `current_position` remains the same, and `read` fails again.

Consider adding a check to ensure progress is made or explicitly handling the error case (e.g., skipping bytes or aborting the current chunk). Or you have to make sure this kind of situation would never happen. (add some comments to explain this if this would never happen)

rattan-core/src/cells/external.rs
```diff
@@ -260,93 +257,151 @@ where
...
+                                                )
+                                                .unwrap_or(&[]),
+                                        ) {
+                                            entry.tcp_entry.seq = tcp_hdr.sequence_number();
+                                            entry.tcp_entry.ack = tcp_hdr.acknowledgment_number();
+                                            entry.tcp_entry.flags =
+                                                *tcp_hdr.slice().get(13).unwrap();
+                                            entry.tcp_entry.dataofs = tcp_hdr.data_offset();
+                                            let entry_bytes = entry.as_bytes().to_owned();
+                                            let _ = tx.send(RattanLogOp::Entry(entry_bytes));
```
BobAnkh: While avoiding blocking the data path is correct, silently swallowing errors might hide issues if the logging thread panics or the channel closes unexpectedly. Consider at least a `tracing::warn!` (rate-limited if possible) or `tracing::debug!` or a counter for dropped log entries.

guide/src/ref/packet-log.md
```diff
@@ -1,37 +1,167 @@
...
+Example usage:
+
+```bash
+python scripts/log_converter.py ./packet.rtl ./output
+```
+
+The paths to input and output files are the same as those for `rattan convert`.
+
+The behaviour of the python script is basically the same as `rattan convert`,
+except for that the output path prefix is not optional.
```
BobAnkh: We might also make it optional I think, so to make the interfaces of our converter consistent.

guide/src/ref/packet-log.md
```diff
@@ -1,37 +1,167 @@
...
+Generated with [protocol](https://github.com/luismartingarcia/protocol), where:
+
+- `LH` is short for log entry header
+
+`LH.length` denotes the whole length of the log entry file, including the header(2 Bytes) and its payload, in Bytes.
+
+Now there are mutiple possibility of `LH.ty.` that decides how the payload of an
+log entry should be interpreted. They are:
+
+- `LH.ty` = 0, for a General Packet Entry as payload.
```
BobAnkh: You should always keep the expression in the same format. You use `LH.ty.` above, you shouldn't use `LH.ty` here, instead, you should still use `LH.ty.`

guide/src/ref/packet-log.md
```diff
@@ -103,18 +230,164 @@ protocol "tcp.flow_id:32,tcp.seq:32,tcp.ack:32,ip.id:16,ip.frag:16,ip.checksum:1
...
+|                                                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 ```
 
-Where:
+```bash
+protocol "LH.length:12,LH.ty.:4,FLE.length:12,FLE.ty.:4,Flow Id:32,src.ip:32,dst.ip:32,src.port:16,dst.port:16,Base Timestamp:64,Reserved:32,Options on SYN/SYN_ACK:320"
+```
+
+where `FLE` is short for FLow Entry, and the meaning of each field is as follows:
```
BobAnkh: While this is flow entry, but the length and ty is in flow entry header. Each entry we should split into header and body, like log entry have the LH section. So we should use `FLH.length` and `FLH.ty.`

guide/src/ref/packet-log.md
BobAnkh: For each field you defined in this specification, you should follow the conventional expression format, an entry type and a dot and an abbreviation for its field. Like we have `LH.ty.`, it belongs to Log Entry Header and its field is abbreviated as `ty.`, which means the type of this Log Entry. We should always follow this convention in all the specification in thie file.

guide/src/ref/packet-log.md
```diff
@@ -103,18 +230,164 @@ protocol "tcp.flow_id:32,tcp.seq:32,tcp.ack:32,ip.id:16,ip.frag:16,ip.checksum:1
...
+
+- `FLE.length` is the length for the whole flow entry. `FLE.type` could be 0 only for now, denoting a TCP/IPv4 flow.
+- `Flow id`: The identifier of the flow. The high 16 bits are the interface ID where the packet enters Rattan, and the low 16 bits are the incremental flow counter.
+- `Base Timestamp`: The base timestamp in nanoseconds (this unit is different from the `GP.timestamp` field) for this flow.
+- `src.ip`, `dst.ip`, `src.port`, `dst.port`: The source IPv4 address, destination IPv4 address, source TCP port number, description TCP port number.
+- `Options on SYN/SYN_ACK`: As flows are uni-directional here, only one valid of SYN(TCP client side) or SYN_ACK (TCP server side) would exist in a TCP connection. The TCP options fileds in such pakcets are recorded in raw here. This field has a fixed length of 40 Bytes, and zero-padding at the end if the tcp options as less than 40 Bytes. 40 Bytes is the upper limit in TCP protocol: a TCP header can be at most 15 \* 4 = 60 Bytes in length, and the first 20 bytes are used for fixed fields in the TCP header.
```
BobAnkh: Should have a space between word and bracket. `SYN(TCP client side)` should be `SYN (TCP client side)`, like you have done in `SYN(TCP client side)`

BobAnkh: Should use `bytes` not `Bytes`, you are mixing them in this sentence.

guide/src/ref/packet-log.md
```diff
@@ -103,18 +230,164 @@ protocol "tcp.flow_id:32,tcp.seq:32,tcp.ack:32,ip.id:16,ip.frag:16,ip.checksum:1
...
+
+#### Chunk Prologue Entry
+
+A chunk prologue entry is a log entry with `LH.typ` = 15, which is used in raw-mode packet logs.
```
BobAnkh: What is `LH.typ`? We only introduce `LH.ty.` in our spec and the the following block you are showing only have `LH.ty.`

guide/src/ref/packet-log.md
```diff
@@ -103,18 +230,164 @@ protocol "tcp.flow_id:32,tcp.seq:32,tcp.ack:32,ip.id:16,ip.frag:16,ip.checksum:1
...
+```
+
+Where `CH` is short for Chunk Prologue Header, and the meaning of each field is as follows:
+
+- `CH.length` denotes the length for the chunk prologue entry, and `CH.ty` is always 0 for now.
+- `Log Version` is a magic number, and it is `0x20251120` in this specification.
+- `Data Length` is the sum of the length of this Chunk Prologue Entry and all the following contents (raw General Packet Entries for now).
+- `Chunk Length` is the logical length of the current chunk, i.e., it is expected to either meet EOF or to find the next Chunk Prologue, `Chunk Length` bytes since the offset of this log entry.
+
+And the custom payload for the `CH.ty` = 0, which is the only defined variant in this specification, is:
```
BobAnkh: Should be `CPH.ty.` instead of `CP.ty`

BobAnkh: Should reorganize this sentence to make it fluent and direct to understand the meaning

src/main.rs
```diff
@@ -58,6 +71,43 @@ pub struct Arguments {
...
+    )]
+    visualized_trace_path: Option<PathBuf>,
+
+    /// Mode to output the visualied_trace.
+    #[arg(
+        long,
+        requires = "visualized_trace",
+        global = true,
+        value_name = "Mode",
+        default_value = "tsv"
```
BobAnkh: We make default to human json

src/main-xdp.rs
BobAnkh: I have some comments in main.rs, this file you should also change as that.

src/visualized_trace.rs
```diff
@@ -0,0 +1,310 @@
...
+const BW_TRACE_SUFFIX: &str = "_bandwidth";
+const DELAY_TRACE_SUFFIX: &str = "_delay";
+
+#[derive(Clone, Copy, Debug, clap::ValueEnum, PartialEq, Eq, Deserialize, Serialize, Default)]
+pub enum OutputMode {
+    #[default]
+    Csv,
+    Tsv,
+    Json,
+    JsonHumanReadable,
```
BobAnkh: Rename to `HumanJson`, the original name is too long.

src/visualized_trace.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+}
+
+struct VisualizedTraceBuilder {
+    inner: TraceRecord,
+    current: Vec<Option<TracePoint>>,
+    current_time: Option<Duration>,
+}
+impl VisualizedTraceBuilder {
```
BobAnkh: A newline is needed between them.

rattan-log/src/logger/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+    // An encoded entry, which can be directly written into log entry file.
+    Entry(Vec<u8>),
+    // A partical built raw log entry.
+    // 3 parts: (flow_id, raw_entry, raw_header)
+    // Raw header shall be written to the `.raw` file.
+    // After that, the flow_id, and the (offset, len) in the `.raw` where the raw_header was written shall
+    // be used to build the raw log entry.
+    RawEntry(u32, RawLogEntry, Vec<u8>),
+    // A flow.
+    // 3 parts: (flow_id , base_time_ns, flow_desc)
```
BobAnkh: Make it a full sentence. For example, `Represent a flow consists of (flow_id, base_time_ns, flow_desc)`

rattan-log/src/logger/writer.rs
```diff
@@ -0,0 +1,199 @@
...
+            None => {
+                self.base_path.set_extension("flow");
+                let file = MmapStreamWriter::new_truncate(&self.base_path)?;
+                self.flow_file.insert(file)
+            }
+        }
+        .extend_from_slice(entry)?;
+        Ok(())
+    }
+    fn add_raw(&mut self, header: &[u8]) -> Result<(u64, usize)> {
```
BobAnkh: Empty newlines between functions

rattan-log/src/log_entry/entry/tcp_ip_compact.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub ack: u32,
+    pub ip_id: u16,
+    pub ip_frag: u16,
+    pub checksum: u16,
+    pub flags: u8,
+    pub dataofs: u8,
+}
+
+unsafe impl Plain for TCPProtocolEntry {}
+static_assertions::assert_eq_size!(TCPProtocolEntry, [u8; 22]);
```
BobAnkh: Add an empty newline between them

rattan-log/src/log_entry/entry/raw.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Debug, Clone, Copy, BinRead, PartialEq, Eq, Default)]
+#[repr(C, packed(2))]
+pub struct RawEntry {
+    pub flow_index: u16,
+    pub pointer: RelativePointer,
+}
+
+unsafe impl Plain for RawEntry {}
+static_assertions::assert_eq_size!(RawEntry, [u8; 6]);
```
BobAnkh: Add an empty newline between them

rattan-log/src/log_entry/entry/flow_entry.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub dst_ip: u32,
+    pub src_port: u16,
+    pub dst_port: u16,
+    pub base_ts: i64,
+    pub _reserved: u32,
+    pub options: TCPOption,
+}
+
+unsafe impl Plain for TCPFlow {}
+static_assertions::assert_eq_size!(TCPFlow, [u8; 70]);
```
BobAnkh: Add an empty newline between them

guide/src/ref/packet-log.md
```diff
@@ -43,7 +43,7 @@ a subcommand `convert`.
 Example usage:
 
 ```bash
-rattan convert ./packet.rtl [./output]
+rattan convert ./packet.rtl [~/path/to/output]
```
BobAnkh: Let's remove the `~/` and just use `path/to/output`

guide/src/ref/packet-log.md
```diff
@@ -78,20 +78,23 @@ info in IP headers are not recorded at all. Artificial fields include:
...
 
 Example usage:
 
 ```bash
-python scripts/log_converter.py ./packet.rtl ./output
+python scripts/log_converter.py ./packet.rtl ~/path/to/output
 ```
 
-The paths to input and output files are the same as those for `rattan convert`.
+or, use the script as a [uv](https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file) script to automatically handle dependencies.
```
BobAnkh: Quote the whole 'uv script': 

```text
[uv script](https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file)
```

rattan-log/src/logger/mod.rs
```diff
@@ -48,17 +48,17 @@ impl PartialEq for FlowDesc {
...
-    // An encoded entry, which can be directly written into log entry file.
+    /// An encoded entry, which can be directly written into log entry file.
     Entry(Vec<u8>),
-    // A partical built raw log entry.
-    // 3 parts: (flow_id, raw_entry, raw_header)
-    // Raw header shall be written to the `.raw` file.
-    // After that, the flow_id, and the (offset, len) in the `.raw` where the raw_header was written shall
-    // be used to build the raw log entry.
+    /// A partial built raw log entry.
+    /// 3 parts: (flow_id, raw_entry, raw_header)
```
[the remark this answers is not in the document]
BobAnkh: I write it for you according to what you have provided:

```text
/// A partially built raw log entry.
/// 3 parts: (flow_id, raw_entry, raw_header)
/// The raw_header must be written to the `.raw` file; only then can we determine
/// its recorded position (represented as a byte range `[offset, offset + len)`). 
/// We use this information (offset and len), along with the flow_id, to construct the raw log entry.
```

Issue #36, opened 2024-04-10
Automate NAT configuration


Centaurus99:
Allow `rattan-cli` to access the internet.

Lethe10137:
A temporary walk-around on Ubuntu 22.04 :

``` bash
# Allow ip forwarding
echo 1 | sudo tee /proc/sys/net/ipv4/ip_forward
sudo sed -i 's/^#*net.ipv4.ip_forward.*/net.ipv4.ip_forward = 1/' /etc/sysctl.conf
sudo sysctl -p

#setup NAT
sudo iptables -t nat -A POSTROUTING -s 10.0.0.0/8 -o <Your network interface> -j MASQUERADE

#setup FORWARD
sudo iptables -A FORWARD -s 10.0.0.0/8 -j ACCEPT
sudo iptables -A FORWARD -d 10.0.0.0/8 -j ACCEPT
```

PR #152, opened 2025-11-28
feat(cli): add NAT management using nftnl

Centaurus99:
This PR adds auto NAT configuration for the Rattan CLI Compatible mode to allow internet access.

**Implementation & Rationale**

I avoided approaches that invoke external binaries, such as [mahimahi's iptables usage](https://github.com/ravinet/mahimahi/blob/f1346c38c415abfc98626dd82450f49a57033603/src/util/nat.cc) or the [nftables crate](https://crates.io/crates/nftables) (which relies on the `nft` binary). Relying on these binaries would necessitate installing `rattan` with SUID (`chmod u+s`), which conflicts with our desired permission management strategy.

Instead, this implementation utilizes [nftnl](https://crates.io/crates/nftnl) to control nftables directly via Netlink. This allows Rattan to function with specific capabilities (`setcap cap_net_admin`) rather than requiring full root privileges.

**Dependencies**

This PR adds `libnftnl-dev` and `libmnl-dev` as build dependencies.

**Related Issues**

Resolves: #36

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,9 +50,10 @@ tracing = "0.1.37"
 tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 
 [features]
-default = ["first-packet"]
+default = ["first-packet", "nat"]
 first-packet = ["rattan-core/first-packet"]
 http = ["rattan-core/http"]
+nat = ["dep:mnl", "dep:nftnl"]
 packet-dump = ["rattan-core/packet-dump"]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


BobAnkh (commented):
The code itself is good. I wonder can we gate it behind a feature and enable it by default (so that we may not need to handle this in company's large scale old version system)? Also, I see you mention that:

> Inner DNS resolution seems to be failing (with systemd-resolved).

I wonder if this PR can be merged? Or do we need any further fix (maybe we should create a new issue if this problem is not solved in this PR).

Centaurus99:
OK, I will add a `--no-nat` command-line argument, and then the PR should be ready to merge.

Regarding the DNS issue, I will open a new issue after some further investigation.

BobAnkh:
> OK, I will add a `--no-nat` command-line argument, and then the PR should be ready to merge.

I hope to clarify this further to prevent any misunderstandings. I am referring to features at compile time; the dependencies may not be easy to install or manage on some older versions of the kernel/release, so we might need a way to avoid compiling the NAT functionality into Rattan's binary files (of course, it can be compiled in by default), which we doesn't need it in the isolated mode for now.

Centaurus99:
Sorry for the misunderstanding. I will add a `nat` feature that is enabled by default, as well as a `--no-nat` command-line argument.

Issue #136, opened 2025-07-16
Order for Config change


maxime-bruno:
As I was implementing #132, I stumbled on a question.

When a cell updates its config, on which packet does it apply?
  * On all packets, including the one already arrived but which haven't been dequeued yet
  * On packets enqueued after the config change

BobAnkh:
For this issue, it is up to the cell itself to decide. However, for the 'official' cell we have currently implemented, it will affect all packets that have not yet been dequeued from the cell to varying degrees (I mean including those that haven't enqueued). But the extent of this impact may differ for those packets(this is related to config). 

Let me give you an example. For instance, in a bandwidth cell, after changing the bandwidth, it will recalculate the next opportunity to send out a packet, thus having an equal effect on all packets that have not yet been enqueued or dequeued in this cell. On the other hand, if we change the queue length and it becomes less than the number of packets already in the queue, then under our current implementation, none of the packets already in the queue will be discarded while any new packets entering the cell will be dropped.

I’m not sure if my explanation and example clarify your doubts, or if you have any further questions or suggestions regarding our current implementation and design?

maxime-bruno:
Your explanation is quite clear. 
So currently it is an undefined behaviour and each cell behave as it want. And the guide line for default cells is try to apply to all packet not yet dequeued.

This raise an issue, it is not how I designed the `DelayPerPacketCell`. So I'll have to change it.

I don't think there is a perfect approach and the current one is ok. But we could argue that the undefined behaviour is not the best idea, especially since the documentation is not really good at the moment.

An idea could be, following the timestamp idea of #132, that when updating a config, we send a timestamp alongside the config. 
This way the config would start to be applied on this instant. And as such packet that should have been dequeued before this instant, are with the old config, the one in the middle is send accordingly to the proper behaviour (like half old, half new bandwidth), and the one after with the new config.

BobAnkh:
> This raise an issue, it is not how I designed the DelayPerPacketCell. So I'll have to change it.

I think this cell has a middle-ground feel, so I can accept both implementations. However, maybe we need some documentation to explain this behavior? It might be complex for us to modify the delays of all packets within the cell accordingly? For this issue, you can decide whether to make changes or not.

> An idea could be, following the timestamp idea of

[...]

Issue #132, opened 2025-07-10
Suggestion to reduce time shift


maxime-bruno:
Hi,

I noticed that most cells that need the arrival of a packet assigned it when the packet arrives in the cell.
Fundamentally, this is pretty logical.
But as we know, du to context switch, wake up delay and else, timer are not the most precise inducing some timeshift.

I tried to implement a more precise timer (https://github.com/maxime-bruno/rattan/tree/fair-timer) with relative success. I'm using a dichotomy to get close to the wake-up timing and then "active wait" to be the most precise.
But as I need to ensure to never wake up early I managed to be more precise but not that much and I'm not sure it justify the burden of active wait.
The only advantage I see here is a way more regular timing (one can check the benches for it)

So while getting a more precise timer, seems nearly impossible without intensive active wait, I was wondering.
Could the packet trait have an invariant defined as :
  * The method `get_timestamp` returns the instant at which the packet was supposed to arrived at the cell
  * The method `set_timestamp` is used only when creating a packet
  * Two new method are added `delay_by` and `delay_until` for cells to add the delay the cell wanted to add. 

That way part of the error of the timer of a cell will be absorbed by the timer of the next cell, as it will delay it less, believing the packet arrived earlier 

I know that this changes could be considered breaking. But I think the idea is interesting to explore.

BobAnkh:
First, we are not worrying about breaking changes, as long as the necessity can be proved. Though we are about to release our first official version, it is fine to add interfaces to the packet trait. However, we may need a more in-depth discussion with our researchers on this issue.

One thing I want to mention here is that `get_timestamp` and `set_timestamp` are designed for each cell to use independently. So I can agree that we can add two interfaces `set_enter_timestamp` and `get_enter_timestamp` to record the time the outside packets enter Rattan (i.e., the packet "created" by Rattan). We can make every cell use set_timestamp to set the time the packet is supposed to leave it, which should logically be the same time the packet is supposed to arrive at the next cell.

I'm not sure the `delay_by` and `delay_until` should be added. It may be hard for some of the cells to decide on it. This is why we still need some sleep-check loops to drive our packets. If you have any additional thoughts on these two interfaces, please provide further information.

I really appreciate this new timer design, but regarding the changes and implementation in this area, I need to ask the core developers @minhuw and @Centaurus99, who previously worked on the related design and implementation, to share their thoughts.

[...]

Issue #131, opened 2025-07-08
Redesign token bucket and bandwidth cells


EwecaBcD:
Our implementation of token bucket and bandwidth cells now process packet enqueuing prior to packet sending, which leads to dropping packets when a burst of packets arrive. Since the cells **must** enqueue packets before dequeuing and sending them, when there is no space in the queue, though the packet can be sent actually, it will be dropped.
This also make Traffic Policing not possible in the token bucket cell. I implemented a temporary version of this cell only for doing experiments in branch fix-tb, but it needs redesigning.

PR #155, opened 2025-12-15
feat: logical timestamp & token bucket (Rebase of #145)

Lethe10137:
Rebase #145 to #150. 

Resolves: #131 
Resolves: #132 
Resolves: #136

### DIFF
--- a/rattan-core/benches/cells/utils.rs
+++ b/rattan-core/benches/cells/utils.rs
@@ -12,7 +12,8 @@ pub fn create_cell<C: Cell<P>, P: Packet, Config: CellFactory<C>>(
     handle: &Handle,
 ) -> Result<C, Error> {
     let mut cell = config(handle)?;
-    cell.receiver().change_state(2);
+    cell.receiver()
+        .change_state(rattan_core::cells::CellState::Normal);
     Ok(cell)
 }
 
--- a/rattan-core/tests/integration/bandwidth.rs
+++ b/rattan-core/tests/integration/bandwidth.rs
@@ -4,6 +4,7 @@ use std::{collections::HashMap, thread::sleep, time::Duration};
 #[cfg(feature = "serde")]
 use std::{sync::mpsc, time::Instant};
 
+use itertools::Itertools;
 use netem_trace::{
     model::{BwTraceConfig, RepeatedBwPatternConfig, StaticBwConfig},
     Bandwidth, BwTrace,
@@ -35,6 +36,7 @@ use tracing::{info, instrument, span, warn, Level};
 
 #[instrument]
 #[test_log::test]
+#[serial_test::parallel]
 fn test_bandwidth() {
     let mut config = RattanConfig::<StdPacket> {
         env: StdNetEnvConfig {
@@ -195,6 +197,7 @@ fn test_bandwidth() {
 #[cfg(feature = "serde")]
 #[instrument]
 #[test_log::test]
+#[serial_test::parallel]
 fn test_droptail_queue() {
     let mut config = RattanConfig::<StdPacket> {
         env: StdNetEnvConfig {
@@ -365,6 +368,7 @@ fn test_droptail_queue() {
 #[cfg(feature = "serde")]
 #[instrument]
 #[test_log::test]
+#[serial_test::parallel]
 fn test_drophead_queue() {
     let mut config = RattanConfig::<StdPacket> {
         env: StdNetEnvConfig {
@@ -540,6 +544,7 @@ fn test_drophead_queue() {
 #[cfg(feature = "serde")]
 #[instrument]
 #[test_log::test]
+#[serial_test::parallel]
 fn test_codel_queue() {
     let mut config = RattanConfig::<StdPacket> {
         env: StdNetEnvConfig {
@@ -747,6 +752,7 @@ fn test_codel_queue() {
 
 #[instrument]
 #[test_log::test]
+#[serial_test::serial]
 fn test_replay() {
     let mut config = RattanConfig::<StdPacket> {
         env: StdNetEnvConfig {
@@ -812,11 +818,11 @@ fn test_replay() {
         let trace_config = RepeatedBwPatternConfig::new().pattern(vec![
             Box::new(StaticBwConfig {
                 bw: Some(Bandwidth::from_mbps(100)),
-                duration: Some(Duration::from_secs(5)),
+                duration: Some(Duration::from_secs(6)),
             }),
             Box::new(StaticBwConfig {
                 bw: Some(Bandwidth::from_mbps(20)),
-                duration: Some(Duration::from_secs(5)),
+                duration: Some(Duration::from_secs(6)),
             }) as Box<dyn BwTraceConfig>,
         ]);
         control_interface
@@ -831,7 +837,7 @@ fn test_replay() {
             .left_spawn(None, move || {
                 let client_handle = std::process::Command::new("iperf3")
                     .args([
-                        "-c", &right_ip, "-p", "9000", "--cport", "10000", "-t", "10", "-J", "-R",
+                        "-c", &right_ip, "-p", "9000", "--cport", "10000", "-t", "12", "-J", "-R",
                         "-C", "reno",
                     ])
                     .stdout(std::process::Stdio::piped())
@@ -855,9 +861,13 @@ fn test_replay() {
             .captures_iter(&stdout)
             .flat_map(|cap| cap[1].parse::<u64>())
             .step_by(2)
-            .take(10)
+            .take(11)
             .collect::<Vec<_>>();
         let front_5s_bandwidth = bandwidth.drain(0..5).collect::<Vec<_>>();
+
+        // There could be a multi-hundred-ms gap between the initial request by iperf and the actual start point
+        // of trasmission. Thus, skip one sec near the change point of bw.
+        let _ = bandwidth.drain(0..1).take(1);
         let back_5s_bandwidth = bandwidth.drain(0..5).collect::<Vec<_>>();
 
         let front_5s_bitrate =
@@ -876,7 +886,81 @@ fn test_replay() {
             back_5s_bandwidth
         );
 
-        assert!(front_5s_bitrate > 90000000 && front_5s_bitrate < 100000000);
-        assert!(back_5s_bitrate > 18000000 && back_5s_bitrate < 22000000);
+        // Correct for difference between payload length and L2 length
+        let front_5s_target_rate = Bandwidth::from_mbps(100).as_bps() as f64 * 1460.0 / 1514.0;
+        let back_5s_target_rate = Bandwidth::from_mbps(20).as_bps() as f64 * 1460.0 / 1514.0;
+
+        let front_5s_bitrate = front_5s_bitrate as f64;
+        let back_5s_bitrate = back_5s_bitrate as f64;
+
+        assert!(front_5s_bitrate > front_5s_target_rate * 0.97);
+        assert!(front_5s_bitrate < front_5s_target_rate * 1.03);
+        assert!(back_5s_bitrate > back_5s_target_rate * 0.97);
+        assert!(back_5s_bitrate < back_5s_target_rate * 1.03);
     }
+}
+
+#[instrument]
+#[test_log::test]
+#[serial_test::parallel]
+fn test_low_rate() {
+    // cargo run -- link --uplink-bandwidth 4096bps --ping -c 10 10.2.1.1 -s 100 -i 0.3
+    let mut config = RattanConfig::<StdPacket> {
+        env: StdNetEnvConfig {
+            mode: StdNetEnvMode::Isolated,
+            client_cores: vec![1],
+            server_cores: vec![3],
+            ..Default::default()
+        },
+        ..Default::default()
+    };
+
+    config.cells.insert(
+        "up_bw".to_string(),
+        CellBuildConfig::Bw(BwCellBuildConfig::Infinite(BwCellConfig::new(
+            Bandwidth::from_bps(4096),
+            InfiniteQueueConfig::new(),
+            None,
+        ))),
+    );
+
+    config.links = HashMap::from([
+        ("left".to_string(), "up_bw".to_string()),
+        ("up_bw".to_string(), "right".to_string()),
+        ("right".to_string(), "left".to_string()),
+    ]);
+    let mut radix = RattanRadix::<AfPacketDriver>::new(config).unwrap();
+    radix.spawn_rattan().unwrap();
+    radix.start_rattan().unwrap();
+
+    // Wait for AfPacketDriver to be ready
+    std::thread::sleep(std::time::Duration::from_millis(100));
+
+    let _span = span!(Level::INFO, "bandwidth_low_rate").entered();
+    info!("try to ping 128B packets in a 4096bps link");
+    let right_ip = radix.right_ip(1).to_string();
+    let left_handle = radix
+        .left_spawn(None, move || {
+            let handle = std::process::Command::new("ping")
+                .args([&right_ip, "-c", "10", "-i", "0.3", "-s", "100"])
+                .stdout(std::process::Stdio::piped())
+                .spawn()
+                .unwrap();
+            Ok(handle.wait_with_output())
+        })
+        .unwrap();
+    let output = left_handle.join().unwrap().unwrap().unwrap();
+    let stdout = String::from_utf8(output.stdout).unwrap();
+
+    let re = Regex::new(r"time=(\d+)").unwrap();
+    let latency = re
+        .captures_iter(&stdout)
+        .flat_map(|cap| cap[1].parse::<f64>())
+        .collect::<Vec<_>>();
+    info!(?latency);
+
+    // Should all be 250ms.
+    let (&min, &max) = latency.iter().minmax().into_option().unwrap();
+    assert!(min >= 248.0);
+    assert!(max <= 250.0);
 }
--- a/rattan-core/tests/integration/compound.rs
+++ b/rattan-core/tests/integration/compound.rs
@@ -25,6 +25,7 @@ use tracing::{info, instrument, span, warn, Level};
 
 #[instrument]
 #[test_log::test]
+#[serial_test::serial]
 fn test_compound() {
     let mut config = RattanConfig::<StdPacket> {
         env: StdNetEnvConfig {
--- a/rattan-core/tests/integration/delay.rs
+++ b/rattan-core/tests/integration/delay.rs
@@ -16,6 +16,7 @@ use tracing::{info, instrument, span, Level};
 
 #[instrument]
 #[test_log::test]
+#[serial_test::parallel]
 fn test_delay() {
     let mut config = RattanConfig::<StdPacket> {
         env: StdNetEnvConfig {
--- a/rattan-core/tests/integration/loss.rs
+++ b/rattan-core/tests/integration/loss.rs
@@ -16,6 +16,7 @@ use tracing::{info, instrument, span, Level};
 
 #[instrument]
 #[test_log::test]
+#[serial_test::parallel]
 fn test_loss() {
     let mut config = RattanConfig::<StdPacket> {
         env: StdNetEnvConfig {
--- a/rattan-core/tests/integration/multipath.rs
+++ b/rattan-core/tests/integration/multipath.rs
@@ -14,6 +14,7 @@ use tracing::warn;
 
 #[instrument]
 #[test_log::test]
+#[serial_test::parallel]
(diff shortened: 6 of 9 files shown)


BobAnkh:
@Lethe10137 Please rebase the PR first. Ping me when it is ready for review

Lethe10137:
tput_comparsion

By modifying the Delay cell, the  recent commit (3fb0ac3) has better tput than current main branch.

rattan-core/src/cells/mod.rs
```diff
@@ -38,8 +49,36 @@ pub trait Packet: Debug + 'static + Send {
...
+        self.delay_until(timestamp);
+    }
+
+    /// Sets the duration the packet should have been delayed by the cell
+    ///
+    /// Like delay_until this should be the theoretical duration spent in the cell.
+    /// This help to avoid over-delaying packets due to sleep time shift.
+    fn delay_by(&mut self, delay: Duration);
+
+    /// Sets the timestamp at which the packet should have left the cell, if the packet was delayed.
```
BobAnkh: `if the packet is delayed`

rattan-core/src/cells/mod.rs
```diff
@@ -256,6 +313,115 @@ impl Packet for StdPacket {
     }
 }
 
+/// A wrapper around a packet structure to analyse a cell's behaviour.
+///
+/// It stores the packet's creation timestamp, allowing us to inspect how long the packet
+/// has been created, in terms of logical timestamp.
+/// This exists only for test code. Especially for the test of cells that may impose
```
BobAnkh: Add a newline between the comments. Make them into two paragraphs in the doc comment.

rattan-core/src/cells/mod.rs
```diff
@@ -362,3 +527,62 @@ macro_rules! wait_until_started {
...
+    ($state:expr, $packet:expr) => {
+        match $state.load(std::sync::atomic::Ordering::Acquire) {
+            $crate::cells::CellState::Drop => return None,
+            $crate::cells::CellState::PassThrough => return Some($packet),
+            $crate::cells::CellState::Normal => $packet,
+        }
+    };
+}
+
+// For test code only. Converts an Instant (machine time) to a relative time since the
```
BobAnkh: `Converts` -> `Convert`

rattan-core/src/cells/token_bucket/mod.rs
```diff
@@ -204,126 +242,42 @@ where
...
-                            // channel closed
-                            return None;
-                        }
+                    // return None if the channel is closed
+                    let recv_packet = recv_packet?;
+                    self.update_timestamp(recv_packet.get_timestamp());
+
+                    let recv_packet = crate::check_cell_state!(self.state, recv_packet);
+
+                    // drop packet if is too large to pass through a full token bucket
```
BobAnkh: `if is` -> `if it is`

BobAnkh (changes requested):

rattan-core/src/cells/mod.rs
```diff
@@ -362,3 +527,62 @@ macro_rules! wait_until_started {
...
+    pub const fn new(state: CellState) -> Self {
+        Self(AtomicU8::new(state as u8))
+    }
+
+    #[inline]
+    pub fn load(&self, order: Ordering) -> CellState {
+        match self.0.load(order) {
+            0 => CellState::Drop,
+            1 => CellState::PassThrough,
+            2 => CellState::Normal,
```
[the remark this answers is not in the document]
Lethe10137: `TryFromPrimitive` was used instead

rattan-core/src/cells/bandwidth/queue.rs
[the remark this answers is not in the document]
Lethe10137: I implemented this in b4c845099974add221cc6fd65bb2f203a26fcbd6

rattan-core/src/cells/delay.rs
```diff
@@ -83,34 +77,47 @@ where
         // Wait for Start notify if not started yet
         crate::wait_until_started!(self, Start);
 
+        if let Ok(config) = self.config_rx.try_recv() {
+            self.set_config(config);
+        }
```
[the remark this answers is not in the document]
BobAnkh: Can you explain it further and more clearly so that I can understand: what is the difference between delay cell with others (like bandwidth or loss, as examples) that makes the situation you mentioned possible? I mean why packet has the opportunity to directly go out  only in this packet not others (you can compare bw and loss as example)?

Because, from theory, the logic of these cells is the same, so I think this might relate to your implementation details?

rattan-core/src/cells/delay.rs
```diff
@@ -296,60 +281,59 @@ where
...
+                (timestamp + *delay_time).max(self.latest_egress_timestamp)
+            } else {
+                timestamp.max(self.latest_egress_timestamp)
+            };
+
+            let sleep_time = logical_send_time.duration_since(Instant::now());
+
+            if !sleep_time.is_zero() {
+                let _ = self.send_timer.sleep(sleep_time).await;
+            }
```
BobAnkh: It isn't correct that you remove the change_timer: what if the delay trace changes during your sleep? This is what I have discussed with you: your local testing might not extensively explore all cases, so you should logically construct all possible cases to handle.

rattan-core/src/cells/bandwidth/mod.rs
```diff
@@ -160,15 +160,14 @@ where
...
-                                0 => {
-                                    return None;
-                                }
-                                1 => {
-                                    return Some(new_packet);
-                                }
-                                _ => {
-                                    self.packet_queue.enqueue(new_packet);
+                            let new_packet = crate::check_cell_state!(self.state, new_packet);
+                            if let Some(packet_to_drop) = self.packet_queue.enqueue(new_packet){
```
BobAnkh: I do not think we should simply enqueue the packet directly into the queue. We should first dequeue all packets whose logical dequeued timestamp is earlier than the logical enqueued timestamp of this packet, so as to make sure it sees a correct queue.

rattan-core/src/cells/bandwidth/mod.rs
```diff
@@ -549,35 +555,35 @@ where
...
         let mut packet = self.packet_queue.dequeue();
+
         while packet.is_none() {
             // the queue is empty, wait for the next packet
             tokio::select! {
                 biased;
                 Some(config) = self.config_rx.recv() => {
                     self.set_config(config);
                 }
-                _ = self.change_timer.sleep(self.next_change - Instant::now()) => {
```
BobAnkh: We should check for the change_timer as I mentioned above.

BobAnkh (changes requested):

rattan-core/src/cells/bandwidth/mod.rs
Lethe10137: I do not think these two goals can be achieved at the same time.

- (1) Avoid too many packets accumulated in the Queue A.
- (2) Maintain a correct queue as you mentioned in Queue B.

where:
Queue A: The queue in the `self.egress`, or the unbounded channel.
Queue B:  The queue that managed by BwCell. e.g. A DropTail queue.

Consider that there are some packets in the Queue A,  by  `first dequeue all packets whose logical dequeued timestamp is earlier than the logical enqueued timestamp of this packet`, we are actually deferring taking packets out from Queue A, which, would accumulate some packets in Queue A  -- When running a 720Mbps BwCell that is actually forwarding at 620Mbps,  the max length of Queue A is between 100 and 300.



Actually it is the choice of orders in the `tokio::select!` at the start of BwCell, wating for the next_available:

This achieves goal (2)Maintain a correct queue as you mentioned in Queue B, for the following reason:

When Queue B is not empty:
logical enqueue timestamp <= "wall clock time of the enqueue"<  self.next_available  <= "logical dequeue timestamp of the head of queue"

```rust
        // wait until next_available
        loop {
            tokio::select! {
                biased;
                Some(config) = self.config_rx.recv() => {
                    self.set_config(config);
                }
                _ = self.timer.sleep(self.next_available - Instant::now()) => {
                    break;
                }
                // `new_packet` can be None only if `self.egress` is closed.
                // The new_packet's timestamp is ealier than self.next_available.
                new_packet = self.egress.recv() => {
                    let new_packet = crate::check_cell_state!(self.state, new_packet?);
                    if let Some(packet) = self.enqueue_packet(new_packet){
                        return Some(packet);
                    }
                }
            }
        }
```

This achieves goal (1) Avoid too many packets accumulated in the Queue A. Because this loop will not exit until Queue A is empty.

Notice that this allows   `Queue B is not empty` `&&` `"wall clock time of the enqueue" >  self.next_available`,
which breaks the chain of inequation above.

```rust
        // wait until next_available
        loop {
            tokio::select! {
                biased;
                Some(config) = self.config_rx.recv() => {
                    self.set_config(config);
                }
             
                // `new_packet` can be None only if `self.egress` is closed.
                // The new_packet's timestamp is ealier than self.next_available.
                new_packet = self.egress.recv() => {
                    let new_packet = crate::check_cell_state!(self.state, new_packet?);
                    if let Some(packet) = self.enqueue_packet(new_packet){
                        return Some(packet);
                    }
                }

                _ = self.timer.sleep(self.next_available - Instant::now()) => {
                          break;
                }
            }
        }
```

Lethe10137: The current version achieves (1), and, the new commit would ignore (1) and try to achieve (2).

BobAnkh: So first, I don't know why we wanted to achieve the goal 1 "Avoid too many packets accumulated in the Queue A" in the past? Does it matter the performance or correctness?

If these two goals conflict, I agree with you that we opt for goal 2.

Besides, I think we may reconsider the overall structure of how we handle the correct ordering in this situation. Maybe we should let the queue itself also maintain a logical timestamp. I mean, the whole rattan is now similar to NS3 now and we advance in logical timestamp, so that the state of our queue should also advance to that timestamp to make it compatible with the outside cell's mechanism. I think this would make our reuse of queues in other cells much easier. Note that the queue may actually contain several buffers to imitate the 'peek' behavior if needed, that is we can cache some packets in the queue's another buffer and wait for enqueue them into the actual working buffer.

We can discuss this mechanism later to see if this is necessary.

Lethe10137: Following the discussion on Jan 4, I implemented that in 372bfb7

BobAnkh: Ping me when it is ready for a final review @Lethe10137

rattan-core/src/metal/io/af_xdp.rs
```diff
@@ -267,6 +268,13 @@ impl Packet for XDPPacket {
         todo!()
     }
 
+    // For test code only.
```
BobAnkh: Change the comments to documentation comments: `//` -> `///`, as this may appear in our docs, so we should inform users.

rattan-core/src/cells/mod.rs
```diff
@@ -362,3 +529,57 @@ macro_rules! wait_until_started {
...
 pub use crate::core::FIRST_PACKET_INSTANT as TRACE_START_INSTANT;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromPrimitive)]
+#[repr(u8)]
+pub enum CellState {
+    // Drops all packets
+    Drop = 0,
+    // Passes through all packets
+    PassThrough = 1,
+    // Normal operation
```
BobAnkh: Use documentation comments

rattan-core/src/cells/mod.rs
```diff
@@ -362,3 +529,57 @@ macro_rules! wait_until_started {
...
+    pub fn store(&self, state: CellState, order: Ordering) {
+        self.0.store(state as u8, order);
+    }
+}
+
+// Check cell state, and:
+// Automatically handle Drop and Passthrough.
+// Packet is returned, on normal conditions.
+//
+// This is the behaviour for most (except ShadowCell) cells.
```
BobAnkh: Let's make it documentation comments. We should clearly state, currently, which CellState corresponds to which action.

rattan-core/src/cells/loss.rs
```diff
@@ -75,19 +85,13 @@ where
...
-                return Some(packet);
-            }
-            _ => {}
+        // It could be None only if the other end of the channel has closed.
+        let packet = self.egress.recv().await?;
+        while let Ok(config) = self.config_rx.try_recv() {
+            self.set_config(config);
         }
+
+        let packet = crate::check_cell_state!(self.state, packet);
```
BobAnkh: I have a question about the order. Maybe we should first check_cell_state then we update the config.

rattan-core/src/cells/config_timestamp.rs
```diff
@@ -0,0 +1,72 @@
+use std::fmt::Debug;
+
+use tokio::time::Instant;
+
+#[cfg(test)]
+use super::relative_time;
+
+/// Store last and current config.
+#[derive(Debug, Default)]
+pub struct CurrentConfig<C> {
```
BobAnkh: Let's rename it to `TimedConfig`

rattan-core/src/cells/config_timestamp.rs
BobAnkh: Rename the file to `timed_config`

rattan-core/src/cells/delay.rs
```diff
@@ -1,20 +1,21 @@
...
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
-use std::fmt::Debug;
-use std::sync::atomic::AtomicI32;
-use std::sync::Arc;
-use tokio::sync::mpsc;
-use tokio::time::Instant;
+use tokio::{sync::mpsc, time::Instant};
+#[cfg(test)]
+use tracing::trace;
```
BobAnkh: Do not import this. When you use tracing::trace and mark it for test only, we could just write the full import, using `tracing::trace!` instead of `trace!`.

rattan-core/src/cells/delay.rs
```diff
@@ -296,60 +287,78 @@ where
...
+                    if self.next_change <= logical_send_time
+                => {
+                    self.update_delay(self.next_change);
                 }
-                _ = self.send_timer.sleep(self.next_available - Instant::now()) => {
-                    break;
+                _ = self.send_timer.sleep(sleep_time),
+                    if self.next_change > logical_send_time
+                => {
+                    break logical_send_time;
```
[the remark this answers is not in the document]
Lethe10137: Sadly, `cargo fmt` can not handle code inside the `select!` macro.  I formatted it by hand, and this it want I think it should look like.

Lethe10137:
> @Lethe10137 cargo-audit fails, which means your new depencency have [security vulnerability](https://github.com/stack-rs/rattan/actions/runs/20913762278/attempts/2#summary-60226784280), please check which depencency use this.

`mnl` is a direct dependency of rattan, introduced when the `nat` feature was implemented.

rattan-core/src/cells/bandwidth/queue.rs
```diff
@@ -1,18 +1,105 @@
...
+    }
+
+    /// Dequeue a packet from the AQM based on the timestamp.
+    /// The function tries to maintain the queue status at the given timestamp before dequeing a packet,
+    /// by enqueuing any packets that should have been enqueued by that timestamp.
+    ///
+    /// The caller ensures that:
+    ///   1) This function is not called before the wall-clock time of `timestamp`.
+    ///   2) The timestamp should be non-descending.
+    pub fn dequeue_at(&mut self, timestamp: Instant) -> Option<P> {
```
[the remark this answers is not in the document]
Lethe10137: What about just handle this multi-path scenario in the `shadow` cell?
For example, just maintain a `max seen timestamp` (`max` ) in the shadow cell, and make sure that a packet either: updates the `max` value, or being delayed to the `max` value.

rattan-core/src/cells/mod.rs
```diff
@@ -19,6 +27,10 @@ pub mod shadow;
 pub mod spy;
 pub mod token_bucket;
 
+pub use config_timestamp::CurrentConfig;
+
+pub const LARGE_DURATION: Duration = Duration::from_secs(10 * 365 * 24 * 60 * 60);
```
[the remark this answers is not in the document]
Lethe10137: No way. If Duration::MAX is used,  we should check for overflow everywhere. 

In algorithm contest, it is a common practice to use `0x3f3f3f3f` instead of `0x7fffffff` for i32::max for similar reason. I think this is a better practice.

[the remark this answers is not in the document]
BobAnkh: Ok, then we can just keep the current value or what you want to be our `LARGE_DURATION`

rattan-core/src/cells/loss.rs
```diff
@@ -60,11 +59,22 @@ where
     /// How many packets have been lost consecutively
     prev_loss: usize,
     rng: R,
-    state: AtomicI32,
+    state: AtomicCellState,
+    config_rx: mpsc::UnboundedReceiver<LossCellConfig>,
```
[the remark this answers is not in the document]
Lethe10137: No special reason but I tried to keep  the behavior of this cell similar to the other cells. I will change it back.

rattan-core/src/cells/shadow.rs
```diff
@@ -47,12 +46,12 @@ where
         crate::wait_until_started!(self, Start);
 
         match self.state.load(std::sync::atomic::Ordering::Acquire) {
-            0 => None,
+            CellState::Drop => None,
```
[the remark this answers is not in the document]
BobAnkh: Let's hold this for our discussion to see if the behavior should change. If it is changed, let's change to use the macro, if not, let's make documentation comments for the cell to state the situation.

rattan-core/src/cells/token_bucket/bucket.rs
```diff
@@ -0,0 +1,248 @@
...
+        (self.time_to_fillup + send_time <= self.fill_up_time).then_some(Token {
+            bucket: self,
+            send_time,
+        })
+    }
+
+    // Allows maintaining the tokens (in Bytes) in bucket during a config change.
+    // Not used in current token_bucket implementation.
+    #[allow(unused)]
+    pub fn change_config(
```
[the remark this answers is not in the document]
Lethe10137: I tried to maintain the old behavior of that , when there is a config change to the token bucket, a new token bucket is constructed. 

This function is for state migration upon config change. I changed this behaviour in new commits, that indeed we would do state migration and it is not unused.

BobAnkh:
@Lethe10137 Is everything ready for review again?

rattan-core/src/cells/bandwidth/mod.rs
```diff
@@ -150,87 +188,79 @@ where
...
-                            // channel closed
-                            return None;
-                        }
+                // `new_packet` can be None only if `self.egress` is closed.
+                // Do not return None here, as there may be some packets in the AQM.
+                Some(new_packet) = self.egress.recv(), if need_to_receive => {
+                    // XXX: If state change from `Normal` to others, the packets in the AQM may never be sent.
+                    let new_packet = crate::check_cell_state!(self.state, new_packet);
+                    if new_packet.get_timestamp() >= self.next_available{
+                        need_to_receive = false;
```
BobAnkh: Shall this mean `self.next_available >= Instant::now()` and we can shortcut another loop to just enqueue packet and break here?

rattan-core/src/cells/bandwidth/mod.rs
```diff
@@ -150,87 +188,79 @@ where
...
-        while packet.is_none() {
+        // Here, either:
+        //    1) No more packets can be retrieved from egress, or
+        //    2) A packet, that should enter the queue after `self.next_available` is seen.
+        // Thus the `dequeue_at()` sees a correct queue, containing any packet that should
+        // enter the AQM at `self.next_available`.
+
+        let packet_to_send = self.transmitting_packet.take();
+        if let Some(next_packet) = self.packet_queue.dequeue_at(self.next_available) {
+            // Here, self.transmitting_packet can not be None, thus the next_packet would not be dropped.
```
BobAnkh: What do you mean by this? I cannot understand the comment: if 'self.transmitting_packet can not be None', then the next_packet should always be dropped

BobAnkh (changes requested):
I didn't check the token bucket's implementation. @Lethe10137 You have to check if all issues I previously and currently raised for the bandwidth cell is resolved or not affecting the token bucket cell

BobAnkh:
@Lethe10137 Please also add comments to token bucket cell if it has similar structures to bandwidth cell

Lethe10137:
> @Lethe10137 Please also add comments to token bucket cell if it has similar structures to bandwidth cell

I will check this later today.

PR #197, opened 2026-06-09
add RED, ARED and PIE

CepheusC:
# Design Choice
## RED and Adaptive RED
1. Following the Linux kernel design, Adaptive RED is implemented as a mode of RED and can be enabled through the `adaptive` field in `RedQueueConfig`.
2. `pkt_tx_time` is used to compute the number of "virtual packet departures" during an idle period for average queue length decay: `m = idle_time_us / pkt_tx_time`. The upper layer computes this from link bandwidth `C` and average packet size `avpkt` as `pkt_tx_time = avpkt * 8 / C`, then passes it down as a fixed config.

## PIE
The standard PIE algorithm (RFC 8033) updates the drop probability `p` at a fixed interval `t_update` (default 15ms). Rather than spawning a dedicated periodic timer, this implementation drives the update from `enqueue()` by catching up on all missed update intervals using the packet's logical timestamp. The rationale is threefold:
1. **Architectural generality**: Our `PacketQueue` trait defines a generic interface that all AQM algorithms implement. Inserting a timer into `PieQueue` would require `Arc<Mutex<>>` synchronization with the caller's task; placing the timer in the upper-layer loop (e.g., `BwCell`) would couple it to PIE-specific logic and break composability.
2. **Logical time semantics**: This framework uses packet-embedded logical timestamps (via `Packet::get_timestamp()`) rather than wall-clock time. Using the packet's timestamp ensures that `update_drop_probability()` sees the correct time delta regardless of how the simulation is paced relative to real time.
3. **Practical equivalence under load**: With `t_update = 15ms` and typical packet inter-arrival times on the order of microseconds, at least one `enqueue()` call falls within every update interval under load. The `while` loop in `enqueue()` replays all missed `update_drop_probability()` calls in bulk, producing exactly the same state as a timer-driven implementation.


rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,485 @@
+// Adaptive RED Queue Implementation Reference:
+// https://www.icir.org/floyd/papers/adaptiveRed.pdf#:~:text=We%20find%20that%20this%20re-vised%20version%20of%20Adaptive,length%20in%20a%20wide%20variety%20of%20traffic%20scenarios.
```
un-lock-able: Could we remove the `#:~:text=xxx` fragment from the URL? It's only used for brower hightlight, not the main PDF URL. The link can simply be `https://www.icir.org/floyd/papers/adaptiveRed.pdf`

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,485 @@
...
+        Self::new(AdaptiveRedQueueConfig::default())
+    }
+}
+
+impl<P> AdaptiveRedQueue<P>
+where
+   P: Packet,
+{
+    fn update_avg (&mut self) {
+        match self.is_empty() {
```
un-lock-able: This match adds an extra level of indentation but is effectively handling an boolean condition. Could we use an `if` clause and an early return on `false` branch instead?

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,485 @@
...
+                    );
+                    return;
+                },
+                false => {
+                    self.now_bytes += packet_size;
+                    self.queue.push_back(packet);
+                    self.idle_start = None;
+                }
+            }
+        } else {
```
un-lock-able: Use an early return inside `if` branch so that we can avoid the else here.

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,485 @@
...
+            .config
+            .packet_limit
+            .is_none_or(|limit| self.queue.len() < limit)
+            && self.config.byte_limit.is_none_or(|limit| {
+                self.now_bytes + packet.l3_length() + self.config.bw_type.extra_length() <= limit
+            })
+        {
+            let packet_size = packet.l3_length() + self.get_extra_length();
+            
+            match self.should_drop() {
```
un-lock-able: Using an `if-else` here can provide better readbility as this is only an boolean check.

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,485 @@
...
+                queue_len = self.queue.len(),
+                now_bytes = self.now_bytes,
+                header = ?format!("{:X?}", &packet.as_slice()[0..std::cmp::min(56, packet.length())]),
+                "Drop packet(l3_len: {}, extra_len: {}) due to hard limit", packet.l3_length(), self.config.bw_type.extra_length()
+            );
+        }
+    }
+
+    fn dequeue(&mut self) -> Option<P> {
+        match self.queue.pop_front() {
```
un-lock-able: Maybe an `if let` can be cleaner.

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,485 @@
...
+            self.update_max_p();
+            self.latest_max_p_update = now;
+        }
+
+        if self
+            .config
+            .packet_limit
+            .is_none_or(|limit| self.queue.len() < limit)
+            && self.config.byte_limit.is_none_or(|limit| {
+                self.now_bytes + packet.l3_length() + self.config.bw_type.extra_length() <= limit
```
un-lock-able: Since self.get_extra_length() is defined below, could we use it consistently instead of duplicating the logic?

un-lock-able (commented):
Please also remember to fix warnings from cargo clippy (in github actions) and run cargo fmt after modifications.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -217,6 +232,7 @@ where
                 header = ?format!("{:X?}", &packet.as_slice()[0..std::cmp::min(56, packet.length())]),
                 "Drop packet(l3_len: {}, extra_len: {}) due to RED algorithm", packet.l3_length(), self.get_extra_length()
             );
+            #[allow(clippy::needless_return)]
             return;
```
un-lock-able, line 667: Simply remove the return clause, don't suppress clippy warning. There are a few similar problems, please fix them all.

CepheusC: 1. Regarding `clippy::needless_return`: An early return is required when `should_drop()` is true, which triggers the clippy warning. I have refactored the logic to use `if-else` instead.
2. Regarding `clippy::too_many_arguments`: I removed fields from `RedQueueConfig`, `AdaptiveRedQueueConfig` and `PieQueueConfig` struct that use built-in default values and do not require user configuration.

un-lock-able: The `clippy::too_many_arguments` are somewhat acceptable since the `new` method on config is currently mainly used in test codes. 

Regarding the `clippy::needless_return`: what's the clippy's warning and suggestion if we remove the `#[allow(clippy::needless_return)]` line?

BobAnkh: I do think we could improve if the new method contains too many arguments. We could just use some setter patterns to create a config with default parameters and set them with additional calls.

CepheusC: 1. Regarding `clippy::needless_return`: In the initial `match` version, the `return` statement was at the end of a conditional branch of the funtion, which made Clippy consider it redundant. I have now reverted it to an `early return` style (where subsequent statements follow the `return`), which will not trigger the warning.
2. Regarding `clippy::too_many_arguments`: After re-examining the RFCs, linux kernel implementations and the original papers, I found that the number of essential user-configurable parameters is 7 or fewer, meaning the warning is no longer triggered. 

    - If there's a future need for users to tweak other advanced internal parameters, I will add those fields back to *QueueConfig and replace the giant new() method with the Builder/Setter pattern (i.e., initializing via Default::default() and updating specific fields using setter methods).

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,507 @@
...
+        let target_min =
+            self.config.min_th as f64 + 0.4 * (self.config.max_th - self.config.min_th) as f64;
+        let target_max =
+            self.config.min_th as f64 + 0.6 * (self.config.max_th - self.config.min_th) as f64;
+        if self.average_queue_length > target_max && self.config.max_p <= 0.5 {
+            self.config.max_p += (self.config.max_p / 4.0).min(0.01);
+        } else if self.average_queue_length < target_min && self.config.max_p >= 0.01 {
+            self.config.max_p *= 0.9;
+        }
+    }
```
Lethe10137: The ARED paper ask us to constrain $max_p$ is constrained to remain within the range $[0.01, 0.5]$.

Your implementation pushes 0.4999 to about 0.6249. and 0.01 to 0.009.  Actually let's implement that in a simpler way.

```suggestion
    fn update_max_p(&mut self) {
        let target_min =
            self.config.min_th as f64 + 0.4 * (self.config.max_th - self.config.min_th) as f64;
        let target_max =
            self.config.min_th as f64 + 0.6 * (self.config.max_th - self.config.min_th) as f64;
        if self.average_queue_length > target_max {
            self.config.max_p += (self.config.max_p / 4.0).min(0.01);
        } else if self.average_queue_length < target_min {
            self.config.max_p *= 0.9;
        }
        // Clamp to [0.01, 0.5] 
        self.config.max_p = self.config.max_p.clamp(0.01, 0.5);
    }
```

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,507 @@
...
+        if avg >= min_th && avg < max_th {
+            self.count_packet += 1;
+            let p_b = self.config.max_p * (avg - min_th) / (max_th - min_th);
+            let p_a = if self.count_packet as f64 * p_b >= 1.0 {
+                1.0
+            } else {
+                p_b / (1.0 - self.count_packet as f64 * p_b)
+            };
+
+            let rand_val = random_range(0.0..1.0);
```
Lethe10137: This adds non-deterministic behavior.  You can refer to how we used a seeded random generator to make the queue's behavior deterministic.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,424 @@
...
+            if queue.length() == before {
+                drop_count += 1;
+            }
+        }
+
+        // It should drop some packets, but not all of them
+        assert!(
+            drop_count > 0,
+            "Should have dropped some packets probabilistically"
+        );
```
Lethe10137, line 883: It's not good enough to just check "drop some packets, but not all of them".

You could try to calculate actually how many packets (an exact number or a range) is expected to be dropped. Thus, when further modifications on the queue's implementation was made, we can be more confident that, there's no unexpected behavior change by the automated test.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+        }
+
+        // RFC 8033 Section 4.1: Bypass random drop logic to be work conserving
+        let bypass_drop =
+            (self.old_del < self.config.ref_del / 2.0 && self.p < 0.2) || self.queue.len() <= 2;
+        if bypass_drop {
+            return false;
+        }
+
+        let rand_val = random_range(0.0..1.0);
```
Lethe10137: Also,  try to make this deterministic with seeded Random generator.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+        rand_val < self.p
+    }
+
+    fn update_avg_drate(&mut self, pkt_size: usize) {
+        let now = Instant::now();
+        let dq_threshold = 16384; // 16 KB
```
Lethe10137: 16 KiB
```suggestion
        let dq_threshold = 16384; // 16 KiB
```

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+        Self::new(PieQueueConfig::default())
+    }
+}
+
+impl<P> PieQueue<P>
+where
+    P: Packet,
+{
+    fn update_drop_probability(&mut self) {
+        let now = Instant::now();
```
Lethe10137: The `now` should be a parameter passed in, which is the logical timestamp of the packet that triggered this.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+        self.config.packet_limit.is_some_and(|limit| limit == 0)
+            || self.config.byte_limit.is_some_and(|limit| limit == 0)
+    }
+
+    fn enqueue(&mut self, packet: P) {
+        // Simulate time-driven with event-driven approach
+        let interval_update = Instant::now().saturating_duration_since(self.start_update);
+        let t_update = Duration::from_millis(15);
+        if interval_update >= t_update {
+            self.update_drop_probability();
```
Lethe10137, line 705: Pass `packet.get_timestamp()` as an argument. The queue should use this as the "logical time", rather than `Instant::now()`.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+        let mut queue: PieQueue<StdPacket> = PieQueue::new(config.clone());
+
+        // Fake high delay
+        queue.avg_drate = 1000.0;
+        queue.now_bytes = 100000; // delay = 100.0s > ref_del
+
+        std::thread::sleep(Duration::from_millis(15)); // Wait to exceed t_update
+
+        // This enqueue will trigger update_drop_probability()
+        queue.enqueue(create_packet(14));
```
Lethe10137: If the queue were working based on the logical timestamp (the timestamp of the packets),  you could just modify the timestamp of the packets.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+
+        assert!(
+            queue.p > 0.0,
```
Lethe10137: What probability should it be updated to? Please make this more strict.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,424 @@
...
+{
+    fn update_avg(&mut self) {
+        if !self.is_empty() {
+            self.average_queue_length = (1.0 - self.config.w_q) * self.average_queue_length
+                + self.config.w_q * (self.now_bytes as f64);
+            return;
+        }
+
+        if let Some(idle_start) = self.idle_start {
+            let now = Instant::now();
```
Lethe10137: Also, try to use the logical timestamp instead.

Lethe10137 (dismissed):
Please refer to #155 about the "logical timestamp".  Conceptually these time-based queue should try to avoid calling Instant::now(), but refer to the timestamp of the packets, which is when the packet "should" be here.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    fn is_zero_buffer(&self) -> bool {
+        self.config.packet_limit.is_some_and(|limit| limit == 0)
+            || self.config.byte_limit.is_some_and(|limit| limit == 0)
+    }
+
+    fn enqueue(&mut self, packet: P) {
+        // Simulate time-driven with event-driven approach
+        let interval_update = Instant::now().saturating_duration_since(self.start_update);
+        let t_update = Duration::from_millis(15);
```
Lethe10137: Please this update interval configurable and default to be 15ms, according to RFC 8033:

   The update interval, T_UPDATE, is defaulted to be 15 milliseconds.
   It MAY be reduced on high-speed links in order to provide smoother
   response.  The target latency value, QDELAY_REF, SHOULD be set to 15
   milliseconds.

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,507 @@
...
+        self.update_avg();
+
+        let now = Instant::now();
+        if now.saturating_duration_since(self.latest_max_p_update) >= Duration::from_millis(500) {
+            self.update_max_p();
+            self.latest_max_p_update = now;
+        }
+
+        let packet_size = packet.l3_length() + self.get_extra_length();
+        let pass_hard_limit = self
```
EwecaBcD: This `pass_hard_limit` actually means whether current queue length does **NOT** pass the `packet_limit` or `byte_limit`. Please consider to use a better variable name.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+        if interval_update >= t_update {
+            self.update_drop_probability();
+        }
+
+        let packet_size = packet.l3_length() + self.get_extra_length();
+        let pass_hard_limit = self
```
EwecaBcD: Same problem as `pass_hard_limit` in ARED implementation.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+    fn default() -> Self {
+        Self::new(PieQueueConfig::default())
+    }
+}
+
+impl<P> PieQueue<P>
+where
+    P: Packet,
+{
+    fn update_drop_probability(&mut self) {
```
EwecaBcD, line 544: According to RFC 8033 and the PIE implementation in Linux kernel (see https://github.com/torvalds/linux/blob/master/net/sched/sch_pie.c#L425), this update happens **every 15ms periodically**. But in your implementation, it seems to happen only when there is a packet trying to enqueue or dequeue. This may lead to inconsistency between your implementation and the standard implementation.

Consider to add a real timer to trigger this update event, or use some circular update logic in `enqueue` and `dequeue` just like this:
```Rust
let mut interval_update = Instant::now().saturating_duration_since(self.start_update);
let t_update = Duration::from_millis(15);

while interval_update >= t_update {
    self.update_drop_probability(t_update.as_secs_f64() * 1000.0);
    interval_update = Instant::now().saturating_duration_since(self.start_update);
}
```

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+                let dq_int = now.saturating_duration_since(start).as_secs_f64();
+                if dq_int > f64::EPSILON {
+                    let dq_rate = self.dq_count as f64 / dq_int;
+                    if self.avg_drate.abs() < f64::EPSILON {
+                        self.avg_drate = dq_rate;
+                    } else {
+                        let epsilon = 0.125;
+                        self.avg_drate = (1.0 - epsilon) * self.avg_drate + epsilon * dq_rate;
+                    }
+                    self.start_measurement = Some(now);
```
EwecaBcD: It seems that the measurement should be ended here according to the pseudocode in RFC 8033 appendix B:
```
if ( PIE->dq_count_ >= DQ_THRESHOLD) {
    ...
    PIE->in_measurement_ = FALSE;
}
```
Please check your implementation.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+        } else if self.p < 0.0001 {
+            p_increment /= 128.0;
+        } else if self.p < 0.001 {
+            p_increment /= 32.0;
+        } else if self.p < 0.01 {
+            p_increment /= 8.0;
+        } else if self.p < 0.1 {
+            p_increment /= 2.0;
+        }
+        self.p += p_increment;
```
EwecaBcD, line 576: I notice that there is a capping logic in the pseudocode in RFC 8033 appendix B:
```
if (PIE->drop_prob_ >= 0.1 && p > 0.02) {
    p = 0.02;
}
```
But you don't have this in your code. Relative content is in RFC 8033 Section 5.5: Cap Drop Adjustment. Please check if you should add this adjustment to your implementation.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,479 @@
...
+
+    fn should_drop(&self) -> bool {
+        // RFC 8033 Section 4.4: Enqueue packet bypassing random drop if burst_allow > 0
+        if self.burst_allowance > f64::EPSILON {
+            return false;
+        }
+
+        // RFC 8033 Section 4.1: Bypass random drop logic to be work conserving
+        let bypass_drop =
+            (self.old_del < self.config.ref_del / 2.0 && self.p < 0.2) || self.queue.len() <= 2;
```
EwecaBcD: In the pseudocode, the `bypass_drop` is relative to the **byte length** of the queue, not the packet length. Please check your implementation.
```
if ( (PIE->qdelay_old_ < QDELAY_REF/2 && PIE->drop_prob_ < 0.2)
     || (queue_.byte_length() <= 2 * MEAN_PKTSIZE) ) {
    return ENQUE;
}
```

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,424 @@
...
+    fn is_zero_buffer(&self) -> bool {
+        self.config.packet_limit.is_some_and(|limit| limit == 0)
+            || self.config.byte_limit.is_some_and(|limit| limit == 0)
+    }
+
+    fn enqueue(&mut self, packet: P) {
+        self.update_avg();
+
+        let packet_size = packet.l3_length() + self.get_extra_length();
+        let pass_hard_limit = self
```
EwecaBcD: Same problem as `pass_hard_limit` in ARED implementation.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,424 @@
...
+
+            let rand_val = random_range(0.0..1.0);
+            if rand_val < p_a {
+                self.count_packet = 0;
+                true
+            } else {
+                false
+            }
+        } else if avg >= max_th {
+            self.count_packet = 0;
```
EwecaBcD: Same problem as your ARED implementation.

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,507 @@
...
+
+            let rand_val = random_range(0.0..1.0);
+            if rand_val < p_a {
+                self.count_packet = 0;
+                true
+            } else {
+                false
+            }
+        } else if avg >= max_th {
+            self.count_packet = 0;
```
EwecaBcD: According to the code in Linux kernel (see https://github.com/torvalds/linux/blob/master/include/net/red.h#L435), the `count_packet` should be set to `-1` when `avg >= max_th`.

rattan-core/src/cells/bandwidth/queue/ared.rs
```diff
@@ -0,0 +1,507 @@
...
+        if !self.is_empty() {
+            self.average_queue_length = (1.0 - self.config.w_q) * self.average_queue_length
+                + self.config.w_q * (self.now_bytes as f64);
+            return;
+        }
+
+        if let Some(idle_start) = self.idle_start {
+            let now = Instant::now();
+            let idle_duration = now.saturating_duration_since(idle_start);
+            let pkt_tx_time = 120.0; // 1500 bytes * 8 / 100Mbps = 120 us
```
EwecaBcD: Why it's hard-coded `100Mbps` here? The bandwidth can be customized to any value.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,424 @@
...
+        if !self.is_empty() {
+            self.average_queue_length = (1.0 - self.config.w_q) * self.average_queue_length
+                + self.config.w_q * (self.now_bytes as f64);
+            return;
+        }
+
+        if let Some(idle_start) = self.idle_start {
+            let now = Instant::now();
+            let idle_duration = now.saturating_duration_since(idle_start);
+            let pkt_tx_time = 120.0; // 1500 bytes * 8 / 100Mbps = 120 us
```
EwecaBcD: Same problem as your ARED implementation.

rattan-core/src/cells/bandwidth/queue/ared.rs
CepheusC: `pkt_tx_time` represents the average transmission time of each packet in the link. Since the `queue` module does not know the information from the upper-layer `bandwidth` module, I use the difference in timestamps between the first and last packets of the queue divided by the number of packets (minus 1) in adaptive RED to estimate the link speed.

CepheusC:
I have finished modifying the code for RED, Adaptive RED, and PIE, and have strictly aligned the logic with both the RFC specifications and the Linux kernel implementation.

There are currently one remaining point that need confirmation:

1. Copilot noted that the current `retain()` implementation removes packets but does not update `now_bytes`. However, I did not find any corresponding `now_bytes` updates inside `retain()` in other algorithms such as DropTail. What is `retain()` supposed to do exactly? Is the current behavior intentional, or should it be modified to update `now_bytes` as packets are removed?

BobAnkh:
cc @un-lock-able @Lethe10137 @EwecaBcD Please reply to the remaining questions to confirm the design.

Lethe10137:
> I have finished modifying the code for RED, Adaptive RED, and PIE, and have strictly aligned the logic with both the RFC specifications and the Linux kernel implementation.
> 
> There are currently one remaining point that need confirmation:
> 
>     1. Copilot noted that the current `retain()` implementation removes packets but does not update `now_bytes`. However, I did not find any corresponding `now_bytes` updates inside `retain()` in other algorithms such as DropTail. What is `retain()` supposed to do exactly? Is the current behavior intentional, or should it be modified to update `now_bytes` as packets are removed?

The `retain` was only used in the TokenBucket , as far as I know
https://github.com/stack-rs/rattan/blame/bb351ee99f83b64cc2c59860925ad926525c9ecf/rattan-core/src/cells/token_bucket/mod.rs#L133-L138

And this was originally introduced by @EwecaBcD .

Centaurus99:
@CepheusC Please rebase onto the latest main branch so that the changes will only be in `rattan-core/src/cells/bandwidth/queue/`. Then we'll do another quick review and prepare to merge.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,597 @@
...
+                + self.config.w_q * (self.now_bytes as f64);
+            return;
+        }
+
+        if let Some(idle_start) = self.idle_start {
+            let now = packet.get_timestamp();
+            let idle_duration = now.saturating_duration_since(idle_start);
+            let m = idle_duration.as_micros() as f64 / self.config.pkt_tx_time;
+            self.average_queue_length *= f64::powf(1.0 - self.config.w_q, m);
+            self.idle_start = Some(now);
```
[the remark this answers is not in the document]
Centaurus99: @CepheusC Please refine the validate method and call it in both `new` and `configure`. If validation fails, it should return an error (for `new`) or discard the changes (for `configure`), and report the error. Same for PIE.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,514 @@
...
+        }
+        let mut interval_update = packet
+            .get_timestamp()
+            .saturating_duration_since(self.start_update.unwrap());
+        while interval_update >= self.config.t_update {
+            self.update_drop_probability();
+            interval_update = packet
+                .get_timestamp()
+                .saturating_duration_since(self.start_update.unwrap());
+        }
```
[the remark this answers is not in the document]
Centaurus99: Proper validation can prevent this issue.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,525 @@
...
+use crate::cells::Packet;
+
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize), serde(default))]
+#[derive(Debug, Clone)]
+pub struct PieQueueConfig {
+    pub packet_limit: Option<usize>,
+    pub byte_limit: Option<usize>,
+    pub ref_del: f64,       // target delay (sec)
+    pub max_burst: f64,     // MAX_BURST (ms)
+    pub t_update: Duration, // update interval
```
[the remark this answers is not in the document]
Centaurus99: @CepheusC Although it compiles without errors, we should use `crate::utils::serde::duration` to provide human-friendly time serde.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,595 @@
...
+    now_bytes: usize,
+    average_queue_length: f64,
+    count_packet: i32,                    // number of packets since last dropping
+    idle_start: Option<Instant>,          // start time of current idle period
+    latest_max_p_update: Option<Instant>, // latest time when max_p was updated (used in adaptive mode), set by first enqueue
+    rng: StdRng,
+}
+
+impl<P> RedQueue<P> {
+    pub fn new(config: RedQueueConfig) -> Result<Self, &'static str> {
```
Centaurus99: `new()` returns `Result` while `impl_bw_cell_into_factory!` expects `new() -> Self`, so RED/PIE can't be registered in `BwCellBuildConfig`/`BwReplayCellBuildConfig`. Change all queues' `new()` to return `Result`, update `impl_bw_cell_into_factory!` (and `impl_bw_replay_cell_into_factory!`) to match, then add RED/PIE to both enums.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,525 @@
...
+    dq_count: usize,                    // departure count (bytes)
+    start_update: Option<Instant>,      // start time of t_update, set by first enqueue
+    start_measurement: Option<Instant>, // Some(Instant) when in a measurement cycle, None when quit
+    avg_drate: f64,
+    burst_allowance: f64,
+    rng: StdRng,
+}
+
+impl<P> PieQueue<P> {
+    pub fn new(config: PieQueueConfig) -> Result<Self, &'static str> {
```
Centaurus99: Same as `red.rs`: resolved once all queues' `new()` return `Result` and the factory macros are updated.

rattan-core/src/cells/bandwidth/queue/pie.rs
Centaurus99: This `Duration` field is missing the project's custom duration (de)serializer. Every other `Duration` config field (e.g. `CoDelQueueConfig::interval`/`target`) uses:
```rust
#[cfg_attr(feature = "serde", serde(with = "crate::utils::serde::duration"))]
```
which serializes to a human-readable string like `"15ms"`. Without it, `t_update` falls back to the default `{ secs, nanos }` representation, making PIE's config format inconsistent with the rest of the codebase. Please add the attribute.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,525 @@
...
+            config,
+            now_bytes: 0,
+            old_del: 0.0,
+            p: 0.0,
+            dq_count: 0,
+            start_update: None,
+            start_measurement: None,
+            avg_drate: 0.0,
+            burst_allowance: max_burst,
+            rng: StdRng::seed_from_u64(42),
```
Centaurus99: Add a `seed` field to `PieQueueConfig` (default 42) and use it here instead of the hardcoded literal. Same for `red.rs:160`.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,595 @@
...
+        debug!(?config, "New RedQueue");
+        Ok(Self {
+            queue: VecDeque::new(),
+            config,
+            now_bytes: 0,
+            average_queue_length: 0.0,
+            count_packet: -1,
+            idle_start: None,
+            latest_max_p_update: None,
+            rng: StdRng::seed_from_u64(42),
```
Centaurus99: Add a `seed` field to `RedQueueConfig` (default 42) and use it here. See `pie.rs`.

Centaurus99 (changes requested):
The RED/ARED/PIE implementations look solid, but a few integration/consistency items should be addressed before merge — see inline comments.

rattan-core/src/cells/bandwidth/queue/red.rs
CepheusC: Changed `RedQueue::new()` and `PieQueue::new()` to return `Self` instead of `Result<Self, &'static str>` , so both can be registered in `impl_bw_cell_into_factory!` / `impl_bw_replay_cell_into_factory!` and added to `BwCellBuildConfig` / `BwReplayCellBuildConfig`.

`validate()` is still called inside `new()`. If it fails, we `panic!` rather than returning `Err` . Does this approach look reasonable?

Centaurus99: I think we should keep `Result<Self, Err>`, change the `new` methods of other existing `Queue` to also return `Result<Self, Err>`, and modify `impl_bw_cell_into_factory!` / `impl_bw_replay_cell_into_factory!` accordingly.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,613 @@
...
+            rng: StdRng::seed_from_u64(seed),
+        })
+    }
+
+    fn configure(&mut self, config: Self::Config) {
+        if let Err(e) = config.validate() {
+            warn!("RedQueue: discard invalid configure: {}", e);
+            return;
+        }
+        self.config = config;
```
Centaurus99, line 609: `configure()` replaces `self.config` (including `seed`) but never re-seeds `self.rng`, so changing `seed` at runtime is silently ignored. The field looks like it controls the RNG, yet it has no effect after a reconfigure.

Suggestion: re-seed only when the seed actually changes, before overwriting the config:
```rust
if config.seed != self.config.seed {
    self.rng = StdRng::seed_from_u64(config.seed);
}
self.config = config;
```

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,542 @@
...
+            rng: StdRng::seed_from_u64(seed),
+        })
+    }
+
+    fn configure(&mut self, config: Self::Config) {
+        if let Err(e) = config.validate() {
+            warn!("PieQueue: discard invalid configure: {}", e);
+            return;
+        }
+        self.config = config;
```
Centaurus99, line 687: Same issue as `RedQueue::configure`: the new `seed` is stored in `self.config` but `self.rng` is never re-seeded, so a seed change at runtime is ignored. Re-seed when `config.seed != self.config.seed` before overwriting the config.

rattan-core/src/cells/bandwidth/queue/pie.rs
```diff
@@ -0,0 +1,542 @@
...
+    #[cfg_attr(feature = "serde", serde(with = "crate::utils::serde::duration"))]
+    pub t_update: Duration, // update interval
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "serde_default")
+    )]
+    pub bw_type: BwType,
+    #[cfg_attr(
+        feature = "serde",
+        serde(default = "default_pie_seed", skip_serializing_if = "serde_default")
```
Centaurus99: `skip_serializing_if = "serde_default"` skips when `seed == 0` (u64's `Default`), but the deserialize default is `default_pie_seed()` = 42. As a result, `seed: 0` is omitted on serialize and round-trips back to 42, silently losing the configured value.

Suggestion: drop `skip_serializing_if` here so `seed` is always serialized (keep `default = "default_pie_seed"` for the missing-field case):
```rust
#[cfg_attr(feature = "serde", serde(default = "default_pie_seed"))]
pub seed: u64,
```

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,613 @@
...
+    pub adaptive: bool, // enable adaptive mode (ARED): max_p is adjusted dynamically
+
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "serde_default")
+    )]
+    pub bw_type: BwType,
+    #[cfg_attr(
+        feature = "serde",
+        serde(default = "default_red_seed", skip_serializing_if = "serde_default")
```
Centaurus99: Same round-trip issue as `PieQueueConfig::seed`: `skip_serializing_if = "serde_default"` skips on `seed == 0`, but the field default is 42, so `seed: 0` serializes away and reads back as 42. Drop `skip_serializing_if` here too.

rattan-core/src/cells/bandwidth/queue/red.rs
```diff
@@ -0,0 +1,613 @@
...
+        if self.config.adaptive {
+            let now = packet.get_timestamp();
+            if self.latest_max_p_update.is_none() {
+                self.latest_max_p_update = Some(now);
+            }
+            if now.saturating_duration_since(self.latest_max_p_update.unwrap())
+                >= Duration::from_millis(500)
+            {
+                self.update_max_p();
+                self.latest_max_p_update = Some(now);
```
Centaurus99: Resetting `latest_max_p_update` to the packet arrival time inflates the ARED interval above 500 ms when arrivals are sparse. A long idle gap then triggers only one adjustment instead of one per elapsed interval. Linux `sch_red.c` instead re-arms on a fixed `jiffies + HZ/2` grid (`red_adaptative_timer`), decoupled from enqueue. Note that PIE in this PR already steps `start_update` on a fixed grid, so RED is also internally inconsistent.

Suggestion: advance on the fixed 500 ms grid rather than to `now`:
```rust
self.latest_max_p_update =
    Some(self.latest_max_p_update.unwrap() + Duration::from_millis(500));
```
(This removes the drift. A full catch-up loop would additionally need to re-decay `average_queue_length` per step to match the kernel's per-timer `qavg`, so that is a bigger change; flagging the drift as the minimal fix.)

Centaurus99 (changes requested):
Overall this is a clean, well-tested PR: it compiles cleanly and all RED/PIE unit tests pass. Requesting changes for three defects around the new `seed` handling and the adaptive-RED update timing.

- `configure()` ignores a changed `seed` (the RNG is never re-seeded) in both RED and PIE.
- `seed` cannot round-trip through serde: `seed: 0` is dropped on serialize and reads back as 42.
- Adaptive RED resets its update reference to the arrival time instead of a fixed 500 ms grid, so the adjustment interval drifts and idle gaps are not caught up (Linux `sch_red.c` uses a fixed periodic timer).

Details inline.

Centaurus99 (approved):
LGTM. @BobAnkh to make the final decision.

BobAnkh (changes requested):
Basically, I agree on the implementation. However, several things need to be checked and confirmed:

- Check the difference of the implementation (I assume you implement it according to the RFC) with kernel's. If any, document them.
- For the naming of the fileds in both queue config struct, I don't know if they are the same as what kernel calls them. If they are different, we should also document that (saying what in the kernel that field corresponds to in documentation comments)
- Since, both PIE queue and RED queue have complex configurations (than our normal queues), we should add documentation comments to both Config struct. Besides that, we can also leave some documentation comments, including the outer documentation comments used at the top of a file or module

cc @Centaurus99 Please also supervise on these things.

BobAnkh (approved):
I'm fine on the docs, and wait for @Centaurus99 to see the difference and squash-merge the PR

PR #198, opened 2026-07-03
refactor(cell): change PacketQueue::dequeue to dequeue_at(ts)

CepheusC:
### Summary

Add an `Option<Instant>` timestamp parameter to `PacketQueue::dequeue()` to support time-aware queue operations. The `CoDel` queue uses this parameter to avoid `Instant::now()` calls, while simple queue implementations (DropHead, DropTail,
Infinite) ignore it with `None`.

### Usage

This change lays the groundwork for PR #197 (redpie branch) by providing the necessary timestamp plumbing in the `PacketQueue` trait. The subsequent PR will make use of this parameter to implement its feature.

### DIFF
--- a/rattan-core/src/cells/bandwidth/queue/codel.rs
+++ b/rattan-core/src/cells/bandwidth/queue/codel.rs
@@ -179,11 +179,11 @@ where
         }
     }
 
-    fn dequeue(&mut self) -> Option<P> {
+    fn dequeue(&mut self, timestamp: Option<Instant>) -> Option<P> {
         match self.queue.pop_front() {
             Some(mut packet) => {
                 self.now_bytes -= packet.l3_length() + self.config.bw_type.extra_length();
-                let now = Instant::now();
+                let now = timestamp.unwrap_or(Instant::now());
                 let drop = self.should_drop(&packet);
                 trace!(
                     drop,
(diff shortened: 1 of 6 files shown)


BobAnkh:
This shall never be an optional argument. And please rename the interface from `dequeue` to `dequeue_at`. @CepheusC 

For simple queues, they can just ignore the value (but we shall still pass the actual value from the cell), as they do not require it. But the semantics of this interface should be fixed to: the cell wants to dequeue a packet at the timestamp, as in our design philosophy, the queue is a passive-driven component.

BobAnkh (approved):
LGTM

BobAnkh:
Wait for extra two approvals (from @Centaurus99 , @un-lock-able or @Lethe10137 ) to merge this.

un-lock-able (approved):
LGTM

rattan-core/src/cells/bandwidth/queue/codel.rs
Centaurus99: The `now` timestamp should be passed in; `Instant::now` should no longer be used.

Centaurus99: Avoid using `Instant::now` in logging as well. The same applies to the logging below.

Centaurus99 (changes requested):

Centaurus99:
In addition, please squash the changes into a single commit and update the PR title accordingly. @CepheusC

BobAnkh:
> In addition, please squash the changes into a single commit and update the PR title accordingly. @CepheusC

We could squash ourselves when merging, so he doesn't need to worry about that and you can have a clear history to review changes. Also, we should try to compact the head into 50 characters instead of spliting into two lines.

Centaurus99 (approved):
LGTM.

PR #206, opened 2026-07-14
refactor(queue): remove retain from PacketQueue trait

Centaurus99:
The trait-level `retain` was buggy in every impl (never updated `now_bytes` after dropping packets) and is only used by TBF's DropTail queue, so maintaining it on the trait isn't worth it. Remove it from the trait and reimplement it as an inherent method on `DropTailQueue<P>` that correctly maintains `now_bytes`.


BobAnkh (approved):
LGTM

PR #207, opened 2026-07-16
ci: reduce self-hosted runner disk I/O

Centaurus99:
Lower disk I/O on the slow self-hosted runners (test + verify):

- **Preinstall LLVM in the runner image** (ansible-built `github-runner`), dropping the per-run `Cache`/`Install LLVM` steps — a ~1 GB local copy, ~8 min on these disks. Jobs use `/usr/lib/llvm-<N>` via `PATH`/`LIBCLANG_PATH` (5.4->14, else->17).
- **`CARGO_PROFILE_RELEASE_DEBUG=line-tables-only`** — release test binaries were ~92% DWARF; cuts write volume, keeps panic line numbers.
- **Drop `max-parallel: 1`** in the test workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name != github.repository
     steps:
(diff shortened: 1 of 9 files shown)


BobAnkh (approved):
So you can just squash all the commits and merge this pr.

PR #208, opened 2026-07-18
refactor(queue): unify Queue::new() to return Result

CepheusC:
This pr makes two changes to the bandwidth queue system:

1. **Unify `Queue::new()` signature** — Change all queue implementations (InfiniteQueue, DropTailQueue, DropHeadQueue, CoDelQueue) to return `Result<Self, Error>` instead of a direct instance.
2. **Update all interfaces related to Queue::new().**


BobAnkh:
@CepheusC Do not include pie and red in this PR.

Centaurus99 (changes requested):
1. Move `new` into `PacketQueue` trait since `impl_bw_cell_into_factory!` macros already assume a fixed `new(config) -> Result<Self, _>` signature per queue.
2. Replace the per-type `From<XxxQueueConfig>` impls with a single impl `TryFrom` on `PacketQueue` trait.
3. @BobAnkh Is `.expect()` in `impl Default for XxxQueue` (e.g. droptail.rs:71) fine to keep, given all four queues actually never fail here?

BobAnkh:
> 1. Move `new` into `PacketQueue` trait since `impl_bw_cell_into_factory!` macros already assume a fixed `new(config) -> Result<Self, _>` signature per queue.
> 2. Replace the per-type `From<XxxQueueConfig>` impls with a single impl `TryFrom` on `PacketQueue` trait.
> 3. @BobAnkh Is `.expect()` in `impl Default for XxxQueue` (e.g. droptail.rs:71) fine to keep, given all four queues actually never fail here?

Yes, it is fine to keep `except` for never-fail condition.

Centaurus99 (approved):
LGTM.

